PR #3, opened 2017-06-27
Materials builder

shyamd:
Basic materials builder added. Bringing this in as a pull request first to get more eyes on it before we merge.

### DIFF
--- a/emmet/tests/test_utils.py
+++ b/emmet/tests/test_utils.py
@@ -1,0 +1,22 @@
+import unittest
+from emmet.utils import get_mongolike, make_mongolike
+
+
+
+class UtilsTests(unittest.TestCase):
+
+    def test_get_mongolike(self):
+        d = {"a": [{"b": 1}, {"c": {"d": 2}}], "e": {"f": {"g": 3}}, "g": 4, "h":[5, 6]}
+
+        self.assertEqual(get_mongolike(d, "g"), 4)
+        self.assertEqual(get_mongolike(d, "e.f.g"), 3)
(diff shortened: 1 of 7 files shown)


mkhorton:
Kind of a trivial change, but reading `MaterialsBuilder` in `materials.py` and trying to understand it, I see this structure:

```python
class MaterialsBuilder:
    def get_items():
    def process_item():
    def match():
    def new_mat():
    def get_origin():
    def update_mat():
    def update_targets():
    def finalize():
```

Just to enhance readability, suggest keeping only the essential builder functions as top-level functions in the builder, and the other functions nested, since they only seem to get called in one place. In other words:

```python
class MaterialsBuilder:
    def get_items():
    def process_item():
        def match():
        def update_mat():
            def get_origin():
        def new_mat():  
    def update_targets():
    def finalize():
```

Also, `get_mongolike()` seems perfectly reasonable, but if we end up having to do a lot of keypath-style searching, it might be easier just to use a keypath library like [dpath](https://github.com/akesterson/dpath-python) which makes it easy to search subkeys in dictionaries etc. but also has some useful methods for merging dictionaries which might be helpful (similar to `recursive_update` but with additional options for e.g. what behaviour you want if there are duplicate keys).

shyamd:
I tried out dpath. It's versatile, but god awful slow. Right now, the builder can generate ~ 100 materials a second on my computer. With dpath, it slows down by nearly 1000x.

mkhorton:
Shame about dpath, 1000x slowdown is quite something(!) Dict lookups by keypath is really better as a language-level feature anyway, it's unfortunate Python doesn't have it.

PR #4, opened 2017-07-12
Initial commit for some documentation (in progress)

mkhorton:
Just some minimal docs for now. Will move them out of README in future if they get too long.

Would it be possible to get write access for the repo? Will update these as I'm writing them, but wanted to write them in the open while everything is still in flux.


dwinston:
Thanks! Yes, we'll want to move to Sphinx eventually, but using README is right for now. I've given you write access @mkhorton

PR #7, opened 2017-07-14
ICSD drone

JPalakapilly:
Added icsd drone.


emmet/borg/icsd_to_mongo.py
```diff
@@ -0,0 +1,255 @@
...
+        newdata = drone.assimilate(drone, path)
+        if newdata:
+            data.append(json.dumps(newdata, cls=MontyEncoder))
+        status['count'] += 1
+        count = status['count']
+        total = status['total']
+        logger.info('{}/{} ({:.2f}%) done'.format(count, total,
+                                                  count / total * 100))
+
+rootpath = '/home/jpalaks/Documents/MP_Work/Scraped Files'
```
dwinston: You can default `rootpath` to the current working directory or something. Also, wrap it and the invocation below in a

    if __name__ == '__main__':

block so it doesn't get invoked when just importing this module.

dwinston:
This is a great start, @JPalakapilly. Please also include basic structure metadata in the output doc at top-level as per [`atomate.utils.utils.get_meta_from_structure`](https://github.com/hackingmaterials/atomate/blob/14006cd705f8c64a8d1565c783d4f9bc17003beb/atomate/utils/utils.py#L132).

JPalakapilly:
i've added the metadata and changed the default filepath

dwinston:
Great, thanks!

PR #10, opened 2017-10-12
Initial topology builder

mkhorton:
Initial topology builder + some minor readme updates.

Some TODOs before merging:

* Need to test on matgen/NERSC, but need to compile critic2 on there first... it's giving me some headaches at the moment

Other TODOs:

* ~~Create a topology_summary builder that aggregates the topology builder~~

* ~~Thinking about a FileIntegrity builder to visit dirs + check for missing/changed files (pending)~~ (see #13)


mkhorton:
I'm getting a weird issue with bader too ... running it manually on a test dir from garden works, but running it inside the builder gives a `Fortran runtime error: Bad real number in item 6791 of list input` ... no idea why, because it's the same binary/same host.

shyamd:
I like the Tasks/Materials methodology.

1.) process_items should not perform any database access. It can be run under MPI and db connections are not transfered.
2.) Since you're breaking it up as tasks and methodology, you might want to have one builder operate on all tasks and just ignore any in which NSW > 0 for instance or ICHAR > 10 which are calcs that move atoms or are nscf. Then a second builder associates these topology 'tasks' with 'materials' and summarizes.

mkhorton:
Ok, that sounds sensible, will update. Thanks!

mkhorton:
I'm going to merge this now -- I believe it's working, but it needs a large-scale test with the other builders. There are probably some bugs in the topology_summary builder to work out. Will require latest version of pymatgen if anyone wants to run this.

PR #13, opened 2017-10-12
File integrity checker

mkhorton:
This hasn't been tested yet, submitting for comment ... thought it might be a good idea to have a builder that alerts us if (a) files go missing and (b) files become corrupted. This PR is just an idea for how that could work.

### DIFF
--- a/emmet/vasp/builders/file_checker.py
+++ b/emmet/vasp/builders/file_checker.py
@@ -1,0 +1,92 @@
+import logging
+import os
+import glob
+import hashlib
+from datetime import datetime
+
+from pymatgen.core.structure import Structure
+
+from maggma.builder import Builder
+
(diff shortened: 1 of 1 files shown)


dwinston:
I think this is a good idea, especially given that we had a recent issue locating a small portion of task files. Is there utility in logging which files appear missing and which appear corrupted, or will our reponse to corrupt/missing files typically be to re-run a task to refresh the directory? If the latter, then your simple logging strategy of "this task id has an issue" will be fine. @tschaume please comment.

tschaume:
We typically re-run the task to refresh the directory since it's more common that an entire directory is missing and not only a few files. Simply logging what's missing would already be a great start to detect potential issues.

shyamd:
We have some other ideas no how files will be treated in maggma/emmet. Closing this for now.

PR #14, opened 2017-10-12
Store task types with task ids in the materials doc

mkhorton:
@shyamd not sure if this needs to be changed anywhere else, but would be really useful to have


shyamd:
Added this in the new variant of the materials builder.

mkhorton:
Great, ty!

PR #16, opened 2017-12-15
abinit phonon builder

gpetretto:
This builder runs anaddb (https://www.abinit.org/sites/default/files/last/users/generated_files/help_anaddb.html)in order to extract the phonon BS and phonon DOS  directly from the database of the results of the calculations. 

TODO - handle possible other sources for the anaddb netcdf files?
     - add more input parameters to tune anaddb calculation?
     - identify the warning for the materials? (e.g. large ASR and CNSR breaking)
     - add a second collection with the phononwebsite format?
     - store the anaddb netcdf output to speedup possible rerunning of the builder?
     - store the PhononBandStructureSymmLine in gridfs?

### DIFF
--- a/emmet/abinit/builders/phonon.py
+++ b/emmet/abinit/builders/phonon.py
@@ -1,0 +1,352 @@
+import logging
+import tempfile
+import gridfs
+import os
+
+from abipy.dfpt.phonons import get_dyn_mat_eigenvec, match_eigenvectors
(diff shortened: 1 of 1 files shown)


dwinston:
Thanks, @gpetretto! I'll pull this in now and we can go through your TODOs as separate issues/PRs.

PR #17, opened 2018-01-12
[WIP] updates for phonon builders

gpetretto:
Work in progress for updates in the phonon builders

- added WebBSPlotter and WebPhononDosVertPlotter
- added dielectric properties to the main builder
- reduced the size of the phononwebsite json


dwinston:
@gpetretto thanks for starting this. could you please address the `codacy-bot` comments?

gpetretto:
Hi, I apparently cannot get the tests to run on travis due to conflicting dependencies. Even moving abipy at the bottom of the requirements list did not solve the problem. in fact, pybtex installs PyYAML>=3.01 (at the moment 3.13) that conflicts with the pinned version of pyyaml in maggma (3.12). The quickest way to fix this seems to update the pyyaml requirements in maggma. Would this be fine with you?

dwinston:
Done. I released a patch to maggma that does not pin any dependencies in its setup.py (only in its requirements.txt, for testing). I also restarted the travis, but now the problem appears to be a pinned version of pymatgen in abipy:
```
error: pymatgen 2018.6.27 is installed but pymatgen==2018.6.11 is required by {'abipy'}
```

gpetretto:
Thanks Donny and sorry for the "spam" due to all the attempts. Now most of the tests related to the phonon builder are passing. I still have one problem though. I have used the `usetex=True` specification to get better looking labels in the band structures, but this causes the test to fail on travis. I tried to disable it before running the test, but it did not seem to work. If you don't have a better workaround this leaves me with two options: removing it from the builder or adding latex to .travis.yml. Which would you prefer?
Alternatively, I have seen that you are using the sumo library for the plot of the electronic band structure. Would using that library be preferable for the phonons as well?

There are also a couple of other tests that are failing (test_builder in emmet.materials.tests.test_site_descriptors.SiteDescriptorsBuilderTest and test_get_all_site_descriptors in emmet.materials.tests.test_site_descriptors.SiteDescriptorsBuilderTest) but those seem related to a recent update in pymatgen rather than to the addition of the abipy dependency. Can you confirm that?

Once the tex problem is fixed I would suggest to proceed with the merge, if you are happy with the current status of the builder.

dwinston:
LaTeX is fine as a build-system dependency. Go ahead and add it to .travis.yml. [Tectonic](https://tectonic-typesetting.github.io/en-US/install.html) seems like a good LaTeX engine.

@shyamd chose sumo for the electronic band structure because co-plotting BS and DOS is nice, and there are some other advantages wrt visual clarity. I don't know if there is as much benefit for phonon plotting, but I'll let @shyamd comment.

I'll check and confirm the `SiteDescriptorsBuilderTest` issues.

gpetretto:
Thanks. If Tectonic doesn't work I think that another option would be to use the same procedure as in [matplotlib](https://github.com/matplotlib/matplotlib/blob/master/.travis.yml):
```
addons:
  apt:
    packages:
      - texlive-fonts-recommended
      - texlive-latex-base
      - texlive-latex-extra
```

dwinston:
Yeah, it seems tectonic isn't working. Go ahead and do that.

PR #18, opened 2018-01-22
Fix substrates builder

dwinston:
Seems like `match["strain"]` raises a `KeyError` if no elastic tensor.


shyamd:
Hmmm yes. That is a bug in pymatgen. The elastic tensor is not required to calculate strain. Let's merge this for now and update later or just wait for the change?

dwinston:
Up to you. I'm not actively running this builder, but I'm in favor of merging now so that the builder is runnable. I just pointed to the code from https://discuss.materialsproject.org/t/how-to-find-the-substrate-for-my-own-structure/967/2 so it'd be nice if it was okay as is.

PR #32, opened 2018-04-09
Property workflow builder

montoyjh:
Adds a PropertyWorkflowBuilder class.  If you have a property collection with task_ids corresponding to a materials collection, this builder will generate workflows that will generate the remaining data.  Generated workflows are tagged with material task_ids such that they aren't duplicated and use bulk_insert_wfs for efficiency.  The main idea is to streamline the creation of production workflows for materials properties other than structure (e. g. bandstructure, elasticity, etc.).

* Takes around 24 hours for the full elastic workflow collection to be built.  It seems like the primary bottleneck here is in bulk workflow insertion, but I'm not sure if there's a way to make it more efficient.
* In order for the workflow to be completely serializable, workflow functions must be specified as a string, which uses atomate's load_class functionality to generate the associated method.  Similarly, the serialization of the LaunchPad is handled separately because it's not MSONable.

Also includes a bit of cleanup to both the mpworks compatibility builder and the elasticity builders to ensure "task_id" is used rather than "material_id".

### DIFF
--- a/emmet/materials/property_workflows.py
+++ b/emmet/materials/property_workflows.py
@@ -52,7 +52,9 @@ class PropertyWorkflowBuilder(Builder):
                 can be a string to be loaded or a custom method.
                 Note that the builder/runner will not be serializable
                 with custom methods.
-            lpad (LaunchPad): fireworks launchpad to use for adding workflows
+            lpad (LaunchPad or dict): fireworks launchpad to use for adding
+                workflows, can either be None (autoloaded), a LaunchPad
+                instance, or a dict from which the LaunchPad will be invoked
             **kwargs (kwargs): kwargs for builder
         """
         self.source = source
@@ -69,7 +71,12 @@ class PropertyWorkflowBuilder(Builder):
             raise ValueError("wf_function must be callable or a string "
                              "corresponding to a loadable method")
         self.material_filter = material_filter
-        self.lpad = lpad or LaunchPad.auto_load()
+        if lpad is None:
+            self.lpad = LaunchPad.auto_load()
+        elif isinstance(lpad, dict):
+            self.lpad = LaunchPad.from_dict(lpad)
+        else:
+            self.lpad = lpad
 
         super().__init__(sources=[source, materials],
                          targets=[], **kwargs)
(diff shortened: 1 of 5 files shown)


montoyjh:
Just FYI, the failed tests are from the site_descriptor_builder.  Looked into it, but wasn't immediately obvious what was going wrong.

dwinston:
@nisse3000 I fixed your site_descriptor builder. You has changed `matminer.featurizers.site. CoordinationNumber.use_weights` in pymatgen to be in `{'none', 'sum', 'effective'}` but the builder still used `True` or `False`. Your test for no weight returns 20 under 'none' versus having returned 18 under False, so I changed the assertion to 20, but you need to verify that this is okay. Also, in future, please test your builder if making changes to modules in pymatgen/matminer on which it depends. Thanks.

nisse3000:
@dwinston, as the Blame view indicates, the change was done by @ardunn.  I ALWAYS test my changes. I also always try to anticipate issues when, for example, matminer or emmet depends on pymatgen and I make changes to the local_env module in pymatgen.  I frequently contact @shyuep in order to know when the next stable version of pymatgen is released so that I can promptly adapt matminer (and emmet). I was not aware of the change in matminer.  Would you expect me to monitor the packages more regularly (every day in this case) or is "acting when a problem occurs" still sufficient?  Thanks.

dwinston:
@nisse3000 I apologize and I appreciate the candor and detail in your response. It is now obvious to me that I was extremely sloppy in my use of attribution (e.g. "you had changed"). I do not expect anyone (you, @ardunn, anyone) to magically know what packages downstream depend on a certain interface. For a mature and highly-depended-on package such as pymatgen, I appreciate e.g. @shyuep's effort to maintain backwards compatibility with deprecation warnings and a timeline for breakage, but I cannot and do not expect this level of rigor for nascent and rapidly evolving libraries such as matminer and emmet. We can collectively only ensure decent test coverage in order to be notified of breaking changes as they occur and fix them. I am glad you had a test in place that noticed the matminer change, and I'm glad it got fixed.

As you know more about the intended operation of the builder, please verify that my changes to it and the test suite were appropriate. Thank you.

PR #35, opened 2018-04-20
Copybuilder bug - tqdm doesn't play nicely with next()

montoyjh:
Somewhat weirdly, tqdm objects aren't iterators, so they throw a type error with next(), which gets called in the Multiprocessor in the runner.  I'm skeptical that what I've done here is the best way to solve the issue, but casting the tqdm object to an iterator in CopyBuilder.get_items() seems to fix the test.

Also, the unittest wasn't working for me initially because there wasn't an index on the "lu" field on the targets.  Added that and a test for runner compatibility.

### DIFF
--- a/emmet/common/copybuilder.py
+++ b/emmet/common/copybuilder.py
@@ -27,7 +27,7 @@ class CopyBuilder(Builder):
         cursor = source.query(criteria=lu_filter,
                               sort=[(source.lu_field, 1)])
         self.logger.info("Will copy {} items".format(cursor.count()))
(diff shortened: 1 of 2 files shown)


dwinston:
I was looking at your `test_runner` test flabbergasted that it passed because it only inserts and copies over docs with `{"v": "old"}`, and then I realized that those last lines should be `assertEqual`, not `assertTrue`, and of course this means my last two lines in the `test_update_targets` are wrong and should be `assertEqual` as well. *facepalm*. Could you please update all to `assertEqual`? Thanks for the extra set of eyes on this!

PR #38, opened 2018-04-25
vasp.materials: separate out StructureMatcher and group_structures

tschaume:
exposing the StructureMatcher and `group_structures` to be used outside of MaterialsBuilder.


tschaume:
obsolete with #39

PR #43, opened 2018-05-10
Copybuilder additions

montoyjh:
Adds non-incremental and global filter functionality for the copybuilder (essentially enables user to copy only documents matching a certain filter criteria or to operate builder in non-incremental mode if a more "global" sync is desired).  Default operation is unchanged.

Also adds a test db cleanup to the tearDown method in the elastic builder that was missing previously.

### DIFF
--- a/emmet/common/copybuilder.py
+++ b/emmet/common/copybuilder.py
@@ -10,20 +10,26 @@ class CopyBuilder(Builder):
     Can override `target.key` field to filter for target document to update.
     """
 
-    def __init__(self, source, target, key=None, **kwargs):
+    def __init__(self, source, target, key=None, query=None,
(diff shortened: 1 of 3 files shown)


dwinston:
Okay. Thank you for retaining prior operation as the default.

I will merge because the one CI failure is unrelated:
```
FAIL: test_builder (test_site_descriptors.SiteDescriptorsBuilderTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/materialsproject/emmet/emmet/materials/tests/test_site_descriptors.py", line 40, in test_builder
    self.assertEqual(len([t for t in sd_builder.get_items()]), 0)
AssertionError: 3 != 0
```
@montoyjh could you please delegate investigation of this?

PR #53, opened 2018-07-26
Requested site-descriptor changes

nisse3000:
Made site-descriptor builder more explicit to avoid problems when adding new near-neighbor-finding classes into local_env in pymatgen; also, adapted unit tests accordingly.


dwinston:
Thanks!

PR #54, opened 2018-07-31
Add GB builder

ucsdlxg:
Add GBbuilder to emmet. We can use Material builder with GB_materials_settings to build materials stores from tasks, and then use GBbuilder to build Gb stores from materials stores.


dwinston:
@mkhorton would you like to take a look at this? We want to ensure your visualizer gets what it needs.

mkhorton:
@dwinston sure thing -- it should be fine, I've been keeping an eye on the `Gb` class and I think it's ok. One thing we could do is pre-build the visualization json itself, I can look into that.

shyamd:
Hi @ucsdlxg ,

I just got a chance to take a closer look at this. I think we need to see the atomate workflow and make sure that gets in before this is merged in. 

The more fundamental issue is that this is not designed for a generic setup, but you're specific setup that is geared around having a separate set of tasks for bulk calcs vs GBs and only running materials project materials. 

Emmet is designed for anyone who runs atomate workflows to be able to build materials and any other property we rely on in MP.  We expect a generic setup to have a single tasks collection that is a mixture of grain boundary as well as bulk calculations, bandstructure calculations, etc. You can use something in the task metadata (SNL) or transformations to identify it as a GB and then use the materials collection to connect it back to a "bulk" structure.

Fundamentally, we should work on getting a workflow into atomate first and then see how this moves forward.

dwinston:
Okay, so this has sat for a while. @ucsdlxg if we want to run this, I need dumps of the two source collections, if those are still relevant. Please confirm this is up-to-date wrt how you generate your output collection.

shyamd:
I'm going to close this for now as the way in which build the GB set needs to be re-thought.

PR #57, opened 2018-08-17
Removed min and max statistics from structure fingerprint

nisse3000:
This was done based on @computron's suggestion. So far, our benchmark tests point to the fact that the most important statistics are the averages and that adding std. dev. only slightly improves the reliability to distinguish different prototype groups/framworks while making the underling fingerprint (vector) double as large. Adding min. and max. further increases the structure fingerprint size considerably by again not adding much information, given our test set (40 prototypes, approx. 6500 materials). However, the min. and max. might be added again in the future if the evidence suggests so.


dwinston:
Sounds good.

PR #59, opened 2018-08-21
Added composition descritpors for similarity toolbox

nisse3000:
Since the site-descriptor builder actually was always also a structure descriptor builder because of the structure fingerprint that was computed and because I added computation of composition descriptors that are required for the MP website's anticipated similarity toolbox, I renamed the builder to "basic_descriptors.py".  Since it is part of the materials module in emmet, it's effectively a basic materials descriptor builder.  I hope that this makes sense to the maintainers, but I'm also happy to discuss any other suggestion.

So, most of this PR is mainly renaming things, and including the 4 or 5 lines for the composition descriptor calculation.


dwinston:
There is currently only one consumer of this builder (MP website similarity toolbox), so it's fine to build all that it needs in one place.

PR #63, opened 2018-09-10
mp website update for propnet and toecs

montoyjh:
A few minor tweaks to the website builder to allow for propnet and TOEC building.

* Adds a propnet kwarg and some functionality to include data from a propnet collection in a web materials build
* Adds some functionality to recursively scrub the MSONable @class and @module from a document if desired.  I'm doing this for now because I don't want the pymatgen interface to the API to deserialize these automatically (I imagine most users won't have propnet installed).
* Adds an additional alias in the add_elastic function to get the TOEC values.

### DIFF
--- a/emmet/common/tests/test_utils.py
+++ b/emmet/common/tests/test_utils.py
@@ -1,0 +1,22 @@
+from unittest import TestCase
+from copy import copy
+from emmet.common.utils import scrub_class_and_module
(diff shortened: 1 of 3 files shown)


montoyjh:
Also built a preliminary propnet collection (just single-element materials) on matgen2.

PR #74, opened 2019-01-28
Add robocrystallographer builder

utf:
Stores the robocrystallographer condensed structure and descriptions to the `robocrys` store.


utf:
So you don't necessarily need to catch errors in the calc method of the builder?

shyamd:
Map builder takes care of that for catastrophic errors:
https://github.com/materialsproject/maggma/blob/7954845e4b07e30d546534f8913e7cfd9d11d9d4/maggma/builders.py#L211-L221

PR #78, opened 2019-02-28
Build structure visualization JSON

mkhorton:
Should work, still need to push crystal toolkit to PyPI, may add some kwargs also.


mkhorton:
Ok, uploaded to pypi. @dwinston this should be appropriate for the grain boundaries if "color_scale" is changed to "grain_label" (will need to add a kwarg to make this easier).

dwinston:
Okay, I'll look into adapting [Simple3DScene](https://github.com/materialsproject/crystaltoolkit/blob/master/src/lib/components/Simple3DScene.js) for the website to test.

mkhorton:
For completeness, the reason `three-full` is a dep of Simple3DScene and not just `three` is:

https://github.com/materialsproject/crystaltoolkit/blob/2a305faab3b08b852abf416a3500c71f39a8982e/src/lib/components/Simple3DScene.js#L83

and

https://github.com/materialsproject/crystaltoolkit/blob/2a305faab3b08b852abf416a3500c71f39a8982e/src/lib/components/Simple3DScene.js#L385

These two add-ons are distributed with `three` on npm in an examples folder ([OrbitControls](https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/OrbitControls.js) and [ConvexGeometry](https://github.com/mrdoob/three.js/blob/dev/examples/js/geometries/ConvexGeometry.js)) but are not ES6, someone has ported OrbitControls to ES6 [here](https://www.npmjs.com/package/orbit-controls-es6) and `three-full` contains all of the examples as ES6 modules also.

Finally, there is an option on the Python side in the new crystal toolkit to explicitly calculate the convex geometries so that `ConvexGeometry` could be made optional in the JS, but this would require a longer database build and also would result in larger scene JSON.

PR #88, opened 2019-03-29
Central repository for global variables

mkhorton:


mkhorton:
@dwinston if you approve of this in principle, can we also get the MAPI route for find_structures to reflect these same tolerances? really confusing behavior at the moment with inconsistency.

dwinston:
Sure. Current instantiation is
```python
m = StructureMatcher(
    ltol=0.2, stol=0.3, angle_tol=5, primitive_cell=True, scale=True,
    attempt_supercell=False, comparator=ElementComparator()
)
```
, which reflects the proposed values. We can make emmet a dep of the web code to
```python
from emmet.magic_numbers import LTOL, STOL, ANGLE_TOL
```
therein.

mkhorton:
Ok, that's weird though because they're the same numbers ... I don't understand why I'm getting multiple mpids for some structures with different space groups

shyamd:
Should we retain the ability to modify these via inputs and leave these "magic" numbers as defaults? Or should you have to change the magic numbers ?

shyamd:
Also @mkhorton , Materials grouping bins by spacegroup first then does Structure matching. If you just structure match, you will find materials with different spacegroups being similar since the positions, angles, etc. all are within tolerance.

mkhorton:
@shyamd ok, what I want to avoid is a situation like this:

```
struct = mpr.get_structure_by_material_id('mp-19311')
mpr.find_structure(struct)  # returns ['mp-19311', 'mp-1178823']
```

First, we should reach consensus on whether this is a problem -- to my mind, it *is* a problem, but I think it would be reasonable to disagree on that. The reason I think it's a problem is that one of those responses from the API is clearly better than the other one given that the space groups *are* substantially different (for all ranges of tolerances, there's clearly an extra mirror plane in one than the other).

Secondly, to fix, we either need to change the find_structure API route, or we need to change how we're grouping materials; the former seems like the option that would give us the least problems. I'm hesitant to restrict to space group though precisely because of some of the benefits fuzzy structure matching gives us.

---

On the magic number front, I don't think there's a harm in having the input args to specify alternatives (maybe you want to test alternatives etc), the important thing is that the defaults are all consistent. What we're missing however is a record of what builders are run, and with what settings, since this isn't stored in the resulting task doc. What I think we need is a robust reporting method, I have some ideas for this, happy to discuss or attempt to implement in the next infrastructure hacking session.

PR #92, opened 2019-05-16
Fix a bug in bond_valence

kmu:
`pymatgen.core.structure.add_oxidation_state_by_site` does not return decorated structure but updates itself.
This patch solves an error happened in `test_bond_valence.py`


mkhorton:
Thanks @kmu !

PR #95, opened 2019-06-04
Initial creation of defects emmet builder

dbroberg:
Initial working version of defects template.


dwinston:
Thanks for sharing this, @dbroberg.

`Builder.process_item` must not do I/O -- that means no `MPRester` calls (in your case, `process_item` calls `get_bulk_gap_data`, which calls `MPRester` ). In fact, there should be no (explicit) use of `MPRester` in an `emmet` builder -- you should connect directly to materials and bandstructures stores just like you connect to a tasks store. `Builder.get_items` should do all data fetching and pass items to `process_item` for "offline" processing prior to updating target stores in `update_targets`.

You can define, for testing purposes, simple stores that call `MPRester` under the hood -- as long as the builder uses only `Store` interfaces to fetch structures and bandstructures.

We will also need at least a sanity-check test (see test files for other builders) in order to merge this.

shyamd:
I'm taking over this builder as @dbroberg is leaving us :(

PR #96, opened 2019-06-08
New molecular builders: molecules, redox, and website

samblau:
qchem.molecules.py: Q-Chem task -> Molecule collection builder
qchem.task_tagger.py: Identifies Q-Chem task doc task type
Both tested in qchem.tests

molecules.redox.py: Molecules collection -> Redox collection builder
molecules.website.py: Molecules + Redox -> Website collection builder
Both tested in molecules.tests

Note that the Website builder is specifically meant for adding to / extending the Electrolyte Genome web app on the Materials Project website. Also note that task_ids in the website builder right now will very likely clash with the molecule IDs currently used by the molecules collection that the website runs on and thus should be changed before attempting to push the latest calculations to the website.


shyamd:
Looks good. Thanks!

PR #109, opened 2019-09-07
Working insertion electrodes builder

jmmshn:
- made things work with deprecated materials and new sandbox tag


jmmshn:
Looks like the h5py install fails on travis-ci?

PR #113, opened 2020-01-10
fixes for electrode builder

jmmshn:
- Added a test file with Bi-Li-O system
- Added tests for main functions in the electrode builder and most of the auxiliary functions
- updated and cleaned up the code for the electrode builder: considerations for non-symmetric and non-commutative structure matcher equalities.


jmmshn:
It looks like the pytest system is failing during setup.

PR #115, opened 2020-04-02
added Models for electrode document

jmmshn:


emmet-core/emmet/core/electrode.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    evolution: float = Field(
+        ..., description="Remaining amount of element .. TODO get better description",
+    )
+
+
+class VoltagePair(BaseModel):
+    """
+    Based on AbstractVoltagePair from pymatgen
```
shyamd, line 35: Do you want this to be a stub so you can just use `AbstractVoltagePair` in the document as both pymatgen object and a pedantic model or do you want a `VoltagePair` document type? If the latter, is this how you want to structure the data?

emmet-core/emmet/core/electrode.py
```diff
@@ -0,0 +1,136 @@
...
+    voltage: float = Field(
+        ..., description="Voltage of voltage pair",
+    )
+
+    mAh: float = Field(
+        ..., description="Energy in mAh",
+    )
+
+    mass_charge: float = Field(
+        ..., description="Mass of charged pair.",
```
shyamd, line 47: I think you mean "charged end"? pair is used sorta ambiguously in these descriptions. It refers to both end members so these properties are for one of those two not the pair?

emmet-core/emmet/core/electrode.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class Electrode(BaseModel):
+    """
+    Based on AbstractElectrode from pymatgen
+    """
+
+    battery_id: str = Field(..., description="The id for this battery document.")
+
+    framework: StructureMetadata = Field(
```
shyamd, line 82: `StructureMetadata` doesn't contain a structure, so you can either have `Electrode` inherit from `StructureMetadata` and choose the charged or discharged end as the reference structure; or you can store an explicit `framework_metdata` key that is `StructureMetadata` and `framework` is of type `Structure`

emmet-core/emmet/core/electrode.py
jmmshn: OK, I can do the former.
I'm not so sure I understand how the stubs work though

jmmshn:
Close for now, putting basic version into API repo first

PR #118, opened 2020-05-28
Add CalcDB based on Maggma Stores

shyamd:
This adds a preliminary implementation of the CalcDB that should handle most use case for `emmet-cli` using Maggma `Store`.

### DIFF
--- a/emmet-cli/requirements.txt
+++ b/emmet-cli/requirements.txt
@@ -14,3 +14,4 @@ slurmpy==0.0.8
 github3.py==1.3.0
 hpsspy==0.5.1
 multiprocessing-logging==0.3.1
+maggma==0.20.0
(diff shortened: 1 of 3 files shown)


emmet-cli/requirements.txt
```diff
@@ -14,3 +14,4 @@ slurmpy==0.0.8
 github3.py==1.3.0
 hpsspy==0.5.1
 multiprocessing-logging==0.3.1
+maggma==0.20.0
```
jmmshn, line 17: Is there a reason this cannot be the newest maggma?  I'm running into install problems and I'm trying to narrow down the cause.

shyamd: requirements.txt should only matter for tests on CI. Since emmet-cli has no tests, this shouldn't matter. You'll want to make sure setup.py has either a `>=` for the version requirement or is set to the latest version of maggma

shyamd:
Should be implemented in atomate v2, so closing for now.

PR #139, opened 2020-11-12
update MaterialsBuilder and ThermoBuilder for SCAN workflow and MP2020 corrections

rkingsbury:
- Update `MaterialsBuilder` to accommodate calculations generated by the new SCAN workflow
 - Update `ThermoBuilder` to use the MP2020 correction scheme.
 - Ensure that `ThermoBuilder` only uses GGA and GGA+U and ignores materials that only have, e.g., SCAN entries

TODO (see last comment below for my thoughts on these)
 - [x] Revert changes to the `explanation` key so that Materials Django doesn't break
 - [x] Decide whether to store `ComputedEntry` or `ComputedStructureEntry` and the fate of the `scan_structure` key. Currently I have retained `scan_structure` and added `r2scan_structure` and `pbesol_structure` keys.
 - [x] Remove lines that set entry.correction=0 from`thermo2.py` and `electrodes.py` (note: this means before we can run ThermoBuilder, we will need to rerun MaterialsBuilder so that the `corrections` key is set materials `.entries` dicts).


emmet/vasp/task_tagger.py
```diff
@@ -105,7 +118,7 @@ def task_type(inputs, include_calc_type=True):
     except Exception:
         functional = "PBE"
 
-    METAGGA_TYPES = {"TPSS", "RTPSS", "M06L", "MBJL", "SCAN", "MS0", "MS1", "MS2"}
+    METAGGA_TYPES = {"TPSS", "RTPSS", "M06L", "MBJL", "SCAN", "R2SCAN", "MS0", "MS1", "MS2"}
```
rkingsbury, line 121: The logic in `task_type` seems brittle to me in that it depends on us defining what functionals to expect. For example, as written it would return 'GGA+U' for a SCAN+U calc, and `GGA` for a calc using PBE potcars with an unknown functional like RSCAN. 

Can / should we re-do this in a way that's more resilient to future changes? For example, can we just call `Vasprun.run_type` from pymatgen?

See https://github.com/materialsproject/pymatgen/pull/1996

emmet/vasp/settings/materials_settings.json
```diff
@@ -348,7 +469,8 @@
   {
     "materials_key": "scan_structure",
```
rkingsbury, line 470: Instead of `scan_structure`, would it make sense to just store `ComputedStructureEntry` instead of `ComputedEntry` in the `.entries` field? I can imagine users wanting to be able to query each structure, even though they will match within a tolerance

rkingsbury: For now I have retained `scan_structure` and also added `r2scan_structure` and `pbesol_structure` keys

emmet/vasp/settings/materials_settings.json
```diff
@@ -334,21 +415,69 @@
...
+      "GGA+U Static": 10,
+      "GGA+U Structure Optimization": 9,
+      "GGA Static": 8,
+      "GGA Structure Optimization": 7,
+      "PBEsol Static": 6,
+      "PBEsol Structure Optimization": 5,
+      "R2SCAN Static": 4,
+      "R2SCAN Structure Optimization": 3,
+      "SCAN Static": 2,
+      "SCAN Structure Optimization": 1
```
rkingsbury, line 458: I added these as alternatives for populating `.structure` because `.structure` is required to validate a materials doc. In the case of Cl2, my R2SCANs were apparently not structure matching to the GGA calcs, resulting in separate materials docs for just the R2SCAN structures. Without the change I made here, the `.structure` field would not be populated in that case, and the material would be invalidated.

rkingsbury: In my latest test built, this change results in ~450 non-deprecated materials that do NOT have a `gga` or `gga_u` entry. Instead, they just have a scan, r2scan, pbesol, or some combination of those.

rkingsbury:
@shyamd ready for you to review. Now that mongo is back up I looked at the output from my latest build and all seems to be in order. A couple things to think about - please let me know your thoughts:

1. Now that I populate `entries.correction` in the materials builder,  there is redundant logic in `thermo2.py` and `electrodes.py` that sets `entry.correction==0`. I think we should leave this in place for now so that `ThermoBuilder` will work on older materials collections that did not populate `entries.correction`. We can remove the logic later.

2. In the future I think I will want to store `ComputedStructureEntry` instead of `ComputedEntry`, but for the time being I have code that relies on the `r2scan_structure` and `scan_structure` keys and I don't want to break it. So I think we can make the change to `ComputedStructureEntry in a future PR.

rkingsbury:
@shyamd as a reminder - you were going to check Materials Django to see what it expects from the `thermo.explanation` key

rkingsbury:
@shyamd unless you see a compelling reason to add `ComputedStructureEntry` to the materials docs now, I'd say let's defer that for later iterations on the builders. For now, dedicated `scan_structure`, `r2scan_structure`, etc. keys meet my needs. 

This is ready for you to review. After Knowhere is back up and we resolve the hashing issue in pymatgen I'll run a few more test builds to make sure nothing unexpected happens with these latest changes.

PR #148, opened 2020-12-17
Structure Group Docs and Builder

jmmshn:
# Created  StructureGroup document

Only groups material documents by the framework structure.  Allow for groups of size 1 so that the full set of ids in the materials system matches before and after grouping.   This information is used to make the builder incremental.


jmmshn:
@shyamd the PR for the structure group itself is ready,  since this is kind of distinct from the actual electrode builder I think we can finish this PR first and I'll add the fixed electrode builder that puts this the electrode document stuff together.

This is failing mypy on the polar.py file for some reason though.

jmmshn:
I'm running into this error in the workflow.
```
Run mypy --namespace-package emmet-core/emmet
emmet-core/emmet/core/polar.py:47: error: Argument 2 to "sum" has incompatible type "Tuple[Tuple[float, float, float], Tuple[float, float, float], Tuple[float, float, float]]"; expected "Union[int, Sequence[int]]"
emmet-core/emmet/core/polar.py:84: error: Argument 2 to "sum" has incompatible type "Tuple[Tuple[float, float, float], Tuple[float, float, float], Tuple[float, float, float]]"; expected "Union[int, Sequence[int]]"
Found 2 errors in 1 file (checked 30 source files)
```

But I'm running into way more errors when on run mypy on my own computer do you know why this is happening?

PR #149, opened 2020-12-17
Electrode Doc

jmmshn:
Bare branch for defining electrode document used in the InsertionElectrode builder.


jmmshn:
@shyamd  this is ready to merge

.pre-commit-config.yaml
```diff
@@ -2,18 +2,26 @@ repos:
...
-  - repo: https://github.com/ambv/black
-    rev: stable
+    - id: check-yaml
+    - id: end-of-file-fixer
+    - id: trailing-whitespace
+    - id: flake8
+      args: [--max-line-length=131]
+
+  - repo: https://github.com/psf/black
+    rev: 19.10b0 # Replace by any tag/version: https://github.com/psf/black/tags
```
shyamd, line 12: Does `stable` not work anymore?

.pre-commit-config.yaml
```diff
@@ -2,18 +2,26 @@ repos:
...
     hooks:
-      - id: isort
-        files: 'emmet-core/*'
+    - id: isort
+      files: 'emmet-core/.*'
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v0.790'
+    hooks:
+      - id: mypy
```
shyamd, line 27: I've always found this to be slow. It's also hard to enforce on a commit-by-commit basis.

emmet-builders/requirements.txt
```diff
@@ -1,3 +1,3 @@
-pymatgen==2020.4.29
-maggma==0.20.0
+pymatgen>=2020.12.18
+maggma>=0.20.0
```
shyamd, line 2: requirements should be pinned to a version so github and dependabot can incrementally check it. If you need a newer version pin it to the version you need.

emmet-core/emmet/core/electrode.py
```diff
@@ -0,0 +1,222 @@
...
+from pymatgen.entries.computed_entries import ComputedEntry
+
+from emmet.stubs import Composition, Structure
+
+
+class WorkingIon(ElementBase):
+    Li = "Li"
+    Ca = "Ca"
+    Mg = "Mg"
+
```
shyamd, line 19: This can just be pymatgen `Element`.

shyamd (changes requested):
Some small changes. Do you need these large data dump file for testing? This shouldn't be a place where you store data as that makes the repo unwieldy.

emmet-core/emmet/core/electrode.py
jmmshn: So I did this with the API in mind since this can be used directly for the drop down ion selection.

shyamd: I would put that at the API level and not limit it here. The filters are separate from the doc model.

.pre-commit-config.yaml
jmmshn: OK removed, I do like it in general though is there a way to locally turn it on?

.pre-commit-config.yaml
```diff
@@ -2,18 +2,26 @@ repos:
...
-    -   id: end-of-file-fixer
-    -   id: trailing-whitespace
-    -   id: flake8
-  - repo: https://github.com/ambv/black
-    rev: stable
+    - id: check-yaml
+    - id: end-of-file-fixer
+    - id: trailing-whitespace
+    - id: flake8
+      args: [--max-line-length=131]
```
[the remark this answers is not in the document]
jmmshn: Changed

.pre-commit-config.yaml
jmmshn: Forgot why this was added but I just tested with stable and it works fine so changed to stable

emmet-builders/requirements.txt
jmmshn: OK

jmmshn:
> This shouldn't be a place where you store data as that makes the repo unwieldy.
I think following the same cases/pattern as pymatgen here makes life easier for future development.
I think it's also possible to grab the test case directly from the pymatgen installed on the system but that seemed hacky.

PR #160, opened 2021-02-13
Restore last_updated in ThermoBuilder

rkingsbury:
Fixes some bugs in queries, clarifies logging messages, and ensures thermo documents have a `last_updated` field.

### DIFF
--- a/emmet/materials/thermo.py
+++ b/emmet/materials/thermo.py
@@ -57,7 +57,7 @@ class ThermoBuilder(Builder):
         # last calculated
         q = dict(self.query)
         q.update(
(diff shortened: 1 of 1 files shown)


rkingsbury:
@shyamd I have not done anything with the IDs yet, because the current implementation appears to work fine for our build pipeline. There are still some challenges when building from combined collections (e.g., my tasks database plus the production one).

Due to the Mongo outage this afternoon I haven't been able to test these changes as extensively as I'd like. I have pulled materials docs from the `materials_180` collection and successfully inserted and updated docs in a `MemoryStore`. So far so good.

PR #167, opened 2021-02-24
Create MPID Type

shyamd:
This creates and implements an MPID type that should fix any issues working with `prefix-int` IDs and `int`s together in one builder.


emmet-core/emmet/core/mpid.py
```diff
@@ -0,0 +1,24 @@
+from dataclasses import dataclass
+from typing import Union
+
+NOTHING = object()
+
+
+class MPID(str):
```
mkhorton, line 7: Usually wary of subclassing builtin types, but usage seems appropriate.

emmet-core/emmet/core/spectrum.py
```diff
@@ -14,7 +15,7 @@ class SpectrumDoc(StructureMetadata):
     metadata on the structure the spectra pertains to
     """
 
-    material_id: str = Field(
+    material_id: Union[MPID, int] = Field(
```
mkhorton, line 18: Why does this need to support int, my understanding of the idea was that ids will be cast to `MPID`?

emmet-core/emmet/core/mpid.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    def __lt__(self, other: Union["MPID", int, str]):
+
+        # Always sort MPIDs before pure integer IDs
+        if isinstance(other, int):
+            return True
+
+        self_parts = self.split("-")
+        other_parts = other.split()
```
mkhorton, line 22: Should `other_parts` split on `-` too?

mkhorton (approved):
Looks good, thanks for soliciting reviews! Left a few notes for clarification.

emmet-core/emmet/core/spectrum.py
shyamd: The MPID type takes care of the string and make its comparable to ints. Since the ID comes from the task_id, which can still be an int, you can still get int material_ids.

emmet-core/emmet/core/mpid.py
shyamd: Good catch! I wonder why the tests didn't catch this. I'll implement a better test firts.

emmet-core/emmet/core/mpid.py
shyamd: Yeah, I was super wary, but this is one of the few situations where it makes sense.

PR #170, opened 2021-03-02
Working Thermo Builder

jmmshn:
- Remove import error from materials builder
- Tested Materials Builder and Structure Group builder
- Materials builder works fine
- Removed sandboxes from Thermo Builder, now breaks during compatibility checks.
- Various fixes to make ThermoBuilder work.
After examining the output for a partial build it looks like the logic of the builder is fine but some E_above hulls are way too high even for atomic solids.  It appears to be a problem with the deprecation of tasks.


emmet-core/emmet/core/vasp/task.py
```diff
@@ -130,6 +130,10 @@ class TaskDocument(StructureMetadata):
         [], description="The 'raw' calculation docs used to assembled this task"
     )
 
+    # def __post_init__(self):
+    #     if self.task_type is None:
+    #         self.is_valid = False
+
```
jmmshn, line 136: I think we should make unrecognized calcs invalid?

emmet-builders/emmet/builders/vasp/materials.py
```diff
@@ -180,6 +182,7 @@ def get_items(self) -> Iterator[List[Dict]]:
         for formula in to_process_forms:
             tasks_query = dict(temp_query)
             tasks_query["formula_pretty"] = formula
+            tasks_query[""]
```
shyamd, line 185:
```suggestion
```

emmet-builders/emmet/builders/vasp/thermo.py
```diff
@@ -99,57 +100,54 @@ def get_items(self) -> Iterator[List[Dict]]:
...
+            #     # only yield maximal subsets so that we can process a equivalent sandbox combinations at a time
+            #     sandbox_entries = [
+            #         entry
+            #         for entry in entries
+            #         if all(
+            #             sandbox in entry.data.get("_sbxn", []) for sandbox in sandboxes
+            #         )
+            #     ]
+            #
+            #     yield sandboxes, sandbox_entries
```
shyamd, line 121:
```suggestion
```

emmet-builders/emmet/builders/vasp/thermo.py
```diff
@@ -99,57 +100,54 @@ def get_items(self) -> Iterator[List[Dict]]:
...
         )
         chemsys = "-".join(elements)
 
-        self.logger.debug(
-            f"Procesing {len(entries)} entries for {chemsys} - {sandboxes}"
-        )
+        self.logger.debug(f"Procesing {len(entries)} entries for {chemsys}")
 
-        material_entries = defaultdict(defaultdict(list))
+        material_entries = defaultdict(lambda: defaultdict(list))
```
shyamd, line 136: Can you clarify why the lambda is needed?

emmet-builders/emmet/builders/vasp/thermo.py
```diff
@@ -178,9 +176,9 @@ def update_targets(self, items):
         # flatten out lists
         items = list(filter(None, chain.from_iterable(items)))
         # check for duplicates within this set
-        items = list(
-            {(v[self.thermo.key], frozenset(v["sandboxes"])): v for v in items}.values()
-        )
+        # items = list(
+        #     {(v[self.thermo.key], frozenset(v["sandboxes"])): v for v in items}.values()
+        # )
```
shyamd, line 181:
```suggestion
```

emmet-core/emmet/core/vasp/task.py
```diff
@@ -152,6 +160,8 @@ def entry(self) -> ComputedEntry:
             "energy": self.output.energy,
             "parameters": {
                 "potcar_spec": self.input.potcar_spec,
+                "is_hubbard": self.input.is_hubbard,
```
shyamd, line 163: Can we "compute" this from the `hubbards` field rather than relying on it being right in the DB?

shyamd (commented):
Overall looks good. Let's not leave dangling code.

emmet-builders/emmet/builders/vasp/thermo.py
jmmshn: I think that's just how `defaultdict` works. 
You need a callable function which returns something and `lambda : x` is just a function that returns x with no inupt.
So if a key is missing it needs to call that function to get the default.
So `defaultdict(lambda : 2)` is how you would get a default value of 2.

emmet-core/emmet/core/vasp/task.py
jmmshn: probably not since the hubbard U parameters and the actual U values are set independent of each other in VASP.  So unless there is some testing of doing ` LDAU=.TRUE.` calculations with zero U values I'm not 100% sure VASP doesn't do some change to the energy.

shyamd: I mean, isn't `is_hubbard  = len(hubbards) > 0 ` ?

emmet-core/emmet/core/vasp/validation.py
```diff
@@ -5,6 +5,7 @@
 from pydantic import BaseModel, Field, PyObject
 from pymatgen.core import Structure
 
+from emmet.builders import EmmetBuildSettings
```
shyamd, line 8:
```suggestion
```

No, don't import this here. There is a settings type for core emmet in `emmet.core.settings` if needed.

shyamd (commented):
Few more review comments

PR #173, opened 2021-03-09
serialize settings

jmmshn:
The current implementations of settings for some builders are not serializable.
This fix makes the default version of the serializable, but some other kind of fix is probably needed.


shyamd:
Dealt with in #174

PR #177, opened 2021-03-16
Updated materials and thermo builder

jmmshn:
- materials_id now uses the smallest possible value to prevent drift
- fixed typo in thermo builder now works, results on 'mp_core_new'
- changed some install settings to allow more flexibility in the requirements.txt
- pegged to pymatgen version to the master branch


emmet-core/emmet/core/vasp/material.py
```diff
@@ -54,6 +54,12 @@ def from_tasks(
             quality_scores: quality scores for various calculation types
             use_statics: Use statics to define a material
         """
+        if task_group == 0:
```
shyamd, line 57: Should there be `len` in there?

jmmshn: Yeah I'll add a test for and update this PR

shyamd, line 57:
```suggestion
        if len(task_group) == 0:
```

PR #179, opened 2021-03-26
deal with NaN in jsanitize

jmmshn:


shyamd:
Is this still needed?

jmmshn:
no it's in the other PR right?

PR #189, opened 2021-05-07
Electronic structure and search components

munrojm:
This PR contains emmet-core and emmet-builder components for `electronic_structure` and `search` data. Both a pydantic model and initial builder are present for `electronic_structure`. Only an initial builder is present for `search`. 

To do:

- [x]  Create a proper document model for `search` and incorporate into builder
- [x]  Fix data chunking and distribution in builders 
- [x]  Add tests


emmet-core/emmet/core/electronic_structure.py
```diff
@@ -0,0 +1,308 @@
...
+        description="Timestamp for when this document was last updated", default_factory=datetime.utcnow
+    )
+
+    @classmethod
+    def from_bsdos(  # type: ignore[override]
+        cls: Type[T],
+        material_id: Union[MPID, int],
+        dos: Dict[Union[MPID, int], CompleteDos],
+        is_gap_direct: bool,
+        is_metal: bool,
```
shyamd, line 114: Confused why these are parameters? Can you add arg description to the docstring?

emmet-core/emmet/core/mpid.py
```diff
@@ -41,7 +41,7 @@ def __eq__(self, other: object):
             return self.parts == MPID(other).parts
 
     def __str__(self):
-        return "-".join(self.parts)
+        return "-".join((str(i) for i in self.parts))
```
shyamd, line 44: Can you add a test that will fail without this bug fix?

emmet-core/emmet/core/electronic_structure.py
munrojm: These are passed as you can't get `is_gap_direct` or `is_metal` from a DOS object. Since that is the preferred source right now for `band_gap`, `cbm` and `vbm`, I have it manually given in order to populate those at the root level of the doc. There is logic in the builder for which line-mode or static calculation is used.

Yes, I will update the docstring.

requirements-testing.txt
```diff
@@ -6,3 +6,4 @@ pydocstyle==5.0.2
 flake8==3.7.9
 mypy==0.770
 mypy-extensions==0.4.3
+seekpath==2.0.1
```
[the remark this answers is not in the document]
munrojm: I have added seekpath to the core and builder requirements.

emmet-core/emmet/core/mpid.py
munrojm: This was failing because `join()` would not work with the second entry in `parts` being an integer. Added a test to check string conversion.

PR #190, opened 2021-05-08
ThermoBuilder changes to work with BVAnalyzer

rkingsbury:
* Update `BondValenceBuilder` to work with latest maggma
* Add `avg_oxidation_states` to the docs generated by `BondValenceBuilder`. This key has the same format as the first element of `Composition.oxi_state_guesses` and is passed to the `Entry.data["oxidation_states"]` key by `ThermoBuilder`. This key, if present, is in turn read by the MP2020 correction scheme
* lint with `black`

### DIFF
--- a/emmet/materials/bond_valence.py
+++ b/emmet/materials/bond_valence.py
@@ -26,13 +26,12 @@ class BondValenceBuilder(MapBuilder):
         super().__init__(
             source=materials,
             target=bond_valence,
-            ufn=self.calc,
             projection=["structure"],
             timeout=14,
             **kwargs
         )
 
-    def calc(self, item):
+    def unary_function(self, item):
         s = Structure.from_dict(item["structure"])
 
         d = {
--- a/emmet/materials/bond_valence.py
+++ b/emmet/materials/bond_valence.py
@@ -15,7 +15,7 @@ MODULE_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)))
 BOND_VALENCE_SCHEMA = os.path.join(MODULE_DIR, "schema", "bond_valence.json")
 
 
-class BondValenceBuilder(MapBuilder):
+class OxidationStateBuilder(MapBuilder):
     """
     Calculate plausible oxidation states from structures.
     """
--- a/emmet/materials/bond_valence.py
+++ b/emmet/materials/bond_valence.py
@@ -18,7 +18,7 @@ class BondValenceBuilder(MapBuilder):
     Calculate plausible oxidation states from structures.
     """
 
-    def __init__(self, materials, bond_valence, **kwargs):
+    def __init__(self, materials, bond_valence, timeout=14, **kwargs):
 
         self.materials = materials
         self.bond_valence = bond_valence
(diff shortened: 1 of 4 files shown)


rkingsbury:
I've tested this successfully in devops #267. The `BVAnalyzer` builder takes a little over an hour to run (which will add to our total build time) and `ThermoBuilder` + MP2020 corrections completes in ~2h 45 minutes. I think my implementation here is probably a little hacky / inelegant so feel free to suggest ways to clean it up if you see opportunities.

mkhorton:
Looks great to me -- honestly I'm happy with it. The oxi builder is so minimal I really don't think we should over think it, if this works that's all we need, and having this collection being built will be really helpful for other builders. If I had any comment it would be a Ctrl+F to replace BondValence with OxidationState, and that's it. Thanks Ryan.

mkhorton:
If you did want to re-structure as a document model, it would look something like this (haven't tested):

```
class OxidationStateDocument(BaseModel):
	
	possible_species: List[str]
	possible_valences: List[float]
	average_oxidation_States: Dict[str, float]
	method: Literal["BVAnalyzer", "oxi_state_guesses"]
	structure: Structure
	
	@classmethod
	def from_structure(cls, structure):
	
		try:
			bva = BVAnalyzer()
			valences = bva.get_valences(structure)
			possible_species = {
				str(Specie(structure[idx].specie, oxidation_state=valence))
				for idx, valence in enumerate(valences)
			}
	
			structure.add_oxidation_state_by_site(valences)
	
			# construct a dict of average oxi_states for use
			# by MP2020 corrections. The format should mirror
			# the output of the first element from Composition.oxi_state_guesses()
			# e.g. {'Li': 1.0, 'O': -2.0}
			s_list=[(t.specie.element, t.specie.oxi_state) for t in structure.sites]
			s_list = sorted(s_list, key=lambda x: x[0])
			oxi_state_dict = {}
			for c,g in groupby(s_list, key=lambda x: x[0]):
				oxi_state_dict[str(c)] = np.mean([e[1] for e in g])
	
			d = {
				"possible_species": list(possible_species),
				"possible_valences": valences,
				"average_oxidation_states": oxi_state_dict,
				"method": "BVAnalyzer",
				"structure": structure,
			}
			
			return cls(**d)
	
		except Exception as e:
			self.logger.error("BVAnalyzer failed with: {}".format(e))
	
			try:
				first_oxi_state_guess = structure.composition.oxi_state_guesses(max_sites=-50)[
					0
				]
				valences = [first_oxi_state_guess[site.species_string] for site in structure]
				possible_species = {
					str(Specie(el, oxidation_state=valence))
					for el, valence in first_oxi_state_guess.items()
				}
	
				structure.add_oxidation_state_by_site(valences)
	
				d = {
					"possible_species": list(possible_species),
					"possible_valences": valences,
					"average_oxidation_states": first_oxi_state_guess,
					"method": "oxi_state_guesses",
					"structure": structure,
				}
				
				return cls(**d)
	
			except Exception as e:
				self.logger.error("Oxidation state guess failed with: {}".format(e))
```

However, honestly, I'm agnostic -- the oxidation state is an intermediate collection, and won't be exposed directly via the API, so I'm a-ok skipping the document model for this one.

mkhorton:
Ah ok, got it. Actually didn’t see what branch we were on 🤦🏻‍♂️

PR #205, opened 2021-06-14
Release Emmet Builders

shyamd:
This PR will enable the automated release of `emmet-builders`.

### DIFF
--- a/emmet-core/emmet/core/search.py
+++ b/emmet-core/emmet/core/search.py
@@ -261,3 +261,135 @@ class SearchDoc(PropertyDoc):
     # Theoretical
 
     theoretical: bool = Field(True, description="Whether the material is theoretical.")
+
+    @classmethod
+    def from_docs(cls, material_id: MPID, **docs: Dict[str, Dict]):
+        """Converts a bunch of search docs into a SearchDoc"""
+        doc = _copy_from_doc(docs)
+
+        # Reshape document for various sub-sections
+        # Electronic Structure + Bandstructure + DOS
+        doc["electronic_structure"] = {
+            k: v for sub_doc in doc["electronic_structure"] for k, v in sub_doc.items()
+        }
+        if "bandstructure" in doc:
+            doc["has_props"].append("bandstructure")
+        if "dos" in doc:
+            doc["has_props"].append("dos")
+        if "calc_id" in doc:
+            doc["es_source_calc_id"] = doc["calc_id"]
+            del doc["calc_id"]
+
+        # Magnetism
+        doc["spin_polarized"] = "magnetism" in doc
+
+        doc["has_props"] = list(set(doc["has_props"]))
+
+        return SearchDoc(material_id=material_id, **doc)
+
+
+# Key mapping
+search_fields: Dict[str, list] = {
+    "materials": [
+        "nsites",
+        "elements",
+        "nelements",
+        "composition",
+        "composition_reduced",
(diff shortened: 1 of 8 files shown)


shyamd:
Jason, can you review the changes to the search builder? This was done to deal with Flake8 complexity issues and I went a bit ... much

emmet-core/emmet/core/search.py
```diff
@@ -261,3 +261,133 @@ class SearchDoc(PropertyDoc):
...
+        """Converts a bunch of search docs into a SearchDoc"""
+        doc = _copy_from_doc(docs)
+
+        # Reshape document for various sub-sections
+        # Electronic Structure + Bandstructure + DOS
+        if "bandstructure" in doc:
+            doc["has_props"].append("bandstructure")
+        if "dos" in doc:
+            doc["has_props"].append("dos")
+        if "calc_id" in doc:
```
munrojm, line 276: `calc_id` should be switched back to `task_id` to match `ElectronicStructureBaseData`

emmet-core/emmet/core/search.py
```diff
@@ -261,3 +261,133 @@ class SearchDoc(PropertyDoc):
...
+        "structure",
+        "deprecated",
+    ],
+    "thermo": [
+        "uncorrected_energy_per_atom",
+        "energy_per_atom",
+        "formation_energy_per_atom",
+        "energy_above_hull",
+        "is_stable",
+        "equillibrium_reaction_energy_per_atom",
```
munrojm, line 312: Should be `equilibrium_...` I believe this is from a typo in `ThermoDoc`

emmet-core/emmet/core/search.py
```diff
@@ -261,3 +261,133 @@ class SearchDoc(PropertyDoc):
...
+    "electronic_structure": [
+        "band_gap",
+        "efermi",
+        "cbm",
+        "vbm",
+        "is_gap_direct",
+        "is_metal",
+        "bandstructure",
+        "dos",
+        "calc_id",
```
munrojm, line 332: `calc_id` should be switched back to `task_id` to match `ElectronicStructureBaseData`

emmet-core/emmet/core/search.py
```diff
@@ -261,3 +261,133 @@ class SearchDoc(PropertyDoc):
...
+        # Magnetism
+        doc["spin_polarized"] = "magnetism" in doc
+
+        doc["has_props"] = list(set(doc["has_props"]))
+
+        return SearchDoc(material_id=material_id, **doc)
+
+
+# Key mapping
+search_fields: Dict[str, list] = {
```
munrojm, line 289: Do we want to come up with some logic to pull fields from the document models instead of listing them here? Would be more robust to changes. Don't know what that might look like, but it might mean the search docs are also more verbose.

munrojm (commented):
Overall, I think the solution looks good to me. Made some minor in-line comments.

emmet-core/emmet/core/search.py
shyamd: Maybe in the future. For now, this should work fine.

emmet-core/emmet/core/search.py
shyamd: The only relevant line in the old builder was: `                    d[id]["es_source_calc_id"] = doc["calc_id"]` Was this a bug then?

PR #215, opened 2021-06-25
Fix Validation

shyamd:
Sigh it seems there are still lots of bugs in validation. This PR is attempting to fix them all
- add static input sets to validation
- only warn not deprecate on kspacing too large
- ensure LDAU checking happens whenever input has U-Values instead of just input_set expects U-values
- fix negative skip bug in max scf gradient checking
- Only checks the SCF gradient when validating input set to ensure this makes sense (IE not for NSCF calcs)

### DIFF
--- a/emmet-builders/emmet/builders/materials/oxidation_states.py
+++ b/emmet-builders/emmet/builders/materials/oxidation_states.py
@@ -37,7 +37,10 @@ class OxidationStatesBuilder(MapBuilder):
 
     def unary_function(self, item):
         structure = Structure.from_dict(item["structure"])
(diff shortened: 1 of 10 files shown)


mkhorton:
> ensure LDAU checking happens whenever input has U-Values instead of just input_set expects U-values

Is this to ensure GGA calcs don't get deprecated?

shyamd:
> > ensure LDAU checking happens whenever input has U-Values instead of just input_set expects U-values
> 
> Is this to ensure GGA calcs don't get deprecated?

The opposite. This is to ensure if the calc should be a GGA, but was run with +U, it gets deprecated. I was only checking one way previously: if input_set wants +U, check if appropriate +U.

PR #236, opened 2021-07-22
Support R2SCAN, SCAN and PBESol tasks in the new build system

rkingsbury:
Various updates and bugfixes necessary to support additional calc types in the new build system. Specifically, this will enable PBEsol, SCAN, and R2SCAN calcs to be included in the validation and materials collections.


rkingsbury:
@munrojm rebased onto `main` so the changes are much clearer now. All tests passing excepts for `docs`. Thanks!

PR #240, opened 2021-08-04
fix bugs in MPID

rkingsbury:
Fixes several bugs in `MPID` to make more robust, especially when building from a collection of task documents that has heterogeneous task ids (e.g., mixture of ints, MPID, prefixed strings)

## Contributor Checklist

- [ x ] I have broken down my PR scope into the following TODO tasks
   -  [ x ] Fix bugs in `MPID.__lt__` when comparing pure integer MPIDs and add more tests to cover edge cases
   -  [ x ] Support coercing number strings e.g. `'1234'` into `int` when instantiating `MPID`
- [ x ] I have run the tests locally and they passed.
- [ x ] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR

### DIFF
--- a/emmet-core/emmet/core/mpid.py
+++ b/emmet-core/emmet/core/mpid.py
@@ -26,7 +26,7 @@ class MPID(str):
             self.parts = val.parts  # type: ignore
             self.string = val.string  # type: ignore
 
-        elif isinstance(val, int) or isinstance(val, str) and val.isnumeric():
+        elif isinstance(val, int) or (isinstance(val, str) and val.isnumeric()):
             self.parts = ("", int(val))
             self.string = str(val)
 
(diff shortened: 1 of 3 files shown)


emmet-core/emmet/core/mpid.py
```diff
@@ -18,8 +26,8 @@ def __init__(self, val: Union["MPID", int, str]):
             self.parts = val.parts  # type: ignore
             self.string = val.string  # type: ignore
 
-        elif isinstance(val, int):
-            self.parts = (None, val)
+        elif isinstance(val, int) or isinstance(val, str) and val.isnumeric():
```
mkhorton, line 29: Could you wrap this in parentheses? Just for readability to distinguish `(A or B) and C` vs `A or (B and C)`

PR #247, opened 2021-08-19
Deprecation data added to PropertyDoc

munrojm:
Deprecated fields added to `PropertyDoc` to allow for deprecation on a property-by-property basis for relevant collections built from `materials`.


shyamd:
Very cool!

PR #250, opened 2021-08-27
Add timeout to provenance builder

shyamd:
This PR adds a timeout to provenance builder as it seems to stall on some materials.


shyamd:
@munrojm Is this still necessary?

munrojm:
@shyamd, nope I fixed the slowness issues with `provenance`.

PR #264, opened 2021-09-09
Fix pre-chunking for materials and thermo builder

munrojm:


mkhorton:
How was this working before?

munrojm:
@mkhorton, it was always run without pre-chunking which is fine. I am now pre-chunking every builder in the automated pipeline and ran into the bug that way.

mkhorton:
Thanks! Just keeping an eye on the PRs, want to make sure I'm following

PR #277, opened 2021-09-30
Add a LMAXMIX check

mkhorton:
@munrojm could you review this? I'm not familiar with the electronic_structure builder. My main concern was which `warnings` field to put the warning into. I think there's also a bunch of duplicated code here.

Key lines are 244-258.

### DIFF
--- a/emmet-builders/emmet/builders/materials/electronic_structure.py
+++ b/emmet-builders/emmet/builders/materials/electronic_structure.py
@@ -20,7 +20,15 @@ from emmet.core.utils import jsanitize
 
 class ElectronicStructureBuilder(Builder):
     def __init__(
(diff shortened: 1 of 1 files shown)


mkhorton:
Note, I see the bug here. Was confused about how data was being passed around for the process_item. Will take a closer look.

munrojm:
I think the only thing you will have to add is making sure `lmaxmix` gets propagated through the function that selects blessed calculations. https://github.com/materialsproject/emmet/blob/047c8c30db73fbd6e91032f581e06145faccfa6f/emmet-builders/emmet/builders/materials/electronic_structure.py#L533 This is to ensure it is in the updated materials doc that gets passed to `process_item` so you can use it in your check on line 245.

munrojm:
I have added a working version of this this to #317.

PR #286, opened 2021-10-20
Magnetism builder updates

munrojm:
This PR updates the magnetism builder. It also fixes chunking for some additional legacy builders.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
           package: ["emmet-core","emmet-builders"]
 
     steps:
-    - uses: actions/checkout@v2.3.4
+    - uses: actions/checkout@v2.3.5
 
     - uses: actions/setup-python@v2.2.2
       with:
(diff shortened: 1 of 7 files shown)


munrojm:
OK. If we build with that `task_id` the only caveat would be that the magnetism collection would not have an entry for every material. This is because `origins` is not populated in deprecated material entries. This is different from the other property collections, but probably won't be an issue.

mkhorton:
Can you clarify? Every material has a thermo doc entry -- why would ever entry not have a magnetism doc entry?

munrojm:
Thermo is also an exception exception. Only docs for non deprecated materials are built. For the other properties, deprecated materials have an entry, but it is indicated as deprecated in the property doc.

munrojm:
Thermo is also an exception. Only docs for non deprecated materials are built is what I meant. For the other properties, deprecated materials have an entry, but it is indicated as deprecated in the property doc.

munrojm:
Thermo is also an exception. Only docs for non deprecated materials are built is what I meant. For the other properties, deprecated materials have an entry, but it is indicated as deprecated in the property doc.

mkhorton:
Ah, understood. I think it's ok if this is only built for non-deprecated materials.

munrojm:
Okay, that sounds good. I will make those changes then.

PR #379, opened 2022-03-15
ChemEnv integration

JaGeo:
*This is a WIP pull request to integrate results from ChemEnv* (@mkhorton)

I have written a code that can determine coordination envionments for cations based on cation-anion interactions with the SimplestChemenvStrategy and its default settings. As a note: I have reused code from the CrystalToolkit implementation.

I still have to write tests and there are still many ToDos mentioned in my code. 

As I have never contributed to emmet, I might need little more help (sorry).

**Questions that are open from my side:** 
 - Do I need to provide code again to determine oxidation states based on a structure? What would be the best way to do this? (Using the OxidationStateDoc?) I would like to have this consistent with OxidationStateDoc
 - Are the current fields okay? Are there ideas for other fields? I could also discard the Molecule-Field
 - Are the fallback options okay in case of an error?
 - Any other suggestions?

Once these questions are solved, I would proceed with writing tests which should be quite fast with the help of the other available tests.


JaGeo:
Thanks. I will fix the tests soon and finish this pull request. Sorry for the delay. 🙂

emmet-core/emmet/core/chemenv.py
```diff
@@ -0,0 +1,271 @@
...
+    "SH:13",
+    "DD:20",
+    "undefined",
+]
+
+
+class ChemEnvDoc(PropertyDoc):
+    """Coordination environments based on cation-anion bonds computed for all unique cations in this structure."""
+
+    property_name = "environment"
```
munrojm, line 96: I would be more specific here, and use something like `coord_environment`.

emmet-core/emmet/core/chemenv.py
```diff
@@ -0,0 +1,271 @@
...
+
+    chemenv_text: List[str] = Field(
+        description="List of text description of coordination environment for unique cationic species in structure."
+    )
+
+    spacegroup: str = Field(
+        description="Space group that was used to determine unique cations"
+    )
+
+    method: str = Field(
```
munrojm, line 123: Is there an Enum that can be used here?

munrojm:
Thanks @JaGeo! Sorry for the delayed response on this. Is the intention to use the added document model in a new builder? Could you clarify what you are using it for locally?

> * Do I need to provide code again to determine oxidation states based on a structure? What would be the best way to do this? (Using the OxidationStateDoc?) I would like to have this consistent with OxidationStateDoc

The answer to this depends on your answers to my previous questions.

> * Are the fallback options okay in case of an error?

I don't think if you only inherit from `PropertyDoc` that you get a `state` field. You will have to explicitly add one if you want one. I think how this part is handled is also dependent on the application. For instance, if it is to be used in a standalone builder, it might not make sense to insert docs into the database for materials where it has failed. 

> * Are the current fields okay? Are there ideas for other fields? I could also discard the Molecule-Field

Fields look fine to me (except for `state`). Again, I guess this depends on the intended use. I think the molecule objects of the different site environments is pretty neat, though. I would keep it around unless it has the potential to blow up the size of the document.

JaGeo:
Thank you very much for your feedback. I will include
I recently met will @mkhorton and he helped clarifying the open questions. I will hopefully submit an adapted version of the doc and a builder for the ChemEnv soon. I will also test it locally with a database precalculated Oxidation States before submitting it.

JaGeo:
I have updated this pull-request (@mkhorton) 

There are still 3 things missing:
- test for the document
- test whether this works for the whole database and in which time (so far, I only tried the first 2000 structures for tests - I have to move to another system for a larger test)
- decision whether we would like to save the StructureEnvironments as well. They will be computed in any case and could allow users to play around with cutoffs on the fly 

I hope the other fields are now much clearer. It really helped me to see the whole workflow for the MP building to understand how this can be done.

Conflicts are only the version numbers at the moment. I guess this can be easily resolved.

JaGeo:
Everything expect for the last point has been resolved. It might influence the size of the DB massively though.

munrojm:
The size of your built database is quite small as is. Even if it was 10x what it is now, it would be fine for us. Have you tried building with the structure environments included?

mkhorton:
I agree, I think it would be fine to save them. And 8 hours seems very reasonable! We will likely run this on a new database release only, since this will be the only time the structures might change, so this will be no burden.

Once this is merged, I guess the next steps would be for us to run a build on our own build servers, and then add the API to the client too, then I can get to work to adding these to the material detail pages.

JaGeo:
Not yet, but I can hopefully try next week. 

Will now at least solve the mypy problems....

JaGeo:
Okay, I think I've fixed the mypy errors, and I'm going to start creating another database. I'll estimate the size of the entire database based on the first 1000 materials and see if it's doable.

The StructureEnvironments have already been included to the document.

JaGeo:
I just finished the first 1000 structures, and we're up to 420 MB. I believe we are at a factor of 50.... What do you think?

munrojm:
Even 50 times might be alright. I’d say anything under 50 GB on disk is worth considering. Would it be possible to build a larger number just to have some more confidence in the scaling factor?

JaGeo:
Size of documents is now a problem. Would need to figure out how to solve this first. Maybe I can do it middle/end of next week. Thanks for you help. Good night

mkhorton:
Ah, darn, a factor of 50x is a lot. Could you offer an example of the JSON of a single structure environment? Do we need to include everything, or can we pare this down somehow?

I will take a look at the sample database you provided too.

And yes, let's defer to next week--happy weekend all!

JaGeo:
I think we should not include the structure environments for now. I also cannot really test the DB creation - even with GridFsDB - since our MongoDB server seems to small for this.

mkhorton:
We have stopped using GridFS for any "large data" for this reason. We currently use an alternative, `S3Store` and run an instance of [MinIO](https://min.io) for MP.

My thought are that we leave it out for now, but if you've already written the code to store them, keep that code around, and we might explore running + storing the environments using S3Store at a later date once we have the rest working.

munrojm:
Is there anything else you would like to do before it is merged @JaGeo?

JaGeo:
I added one more comment that might help to save the structure environments more easily later on. Otherwise, this should be fine from my side now.

munrojm:
Great! Once the tests pass, I will merge and do a minor release. From there, we will work to get it incorporated into the build pipeline and get a test build complete.

mkhorton:
@munrojm once this is merged, should we add some of this info to the `SummaryDoc` too, so that we can enable searching via the website?

munrojm:
@mkhorton, yes definitely.

JaGeo:
Thank you both so much for your help. Let me know if I can help with anything else.

mkhorton:
Thank you too @JaGeo! Excited to get this out to people :)

munrojm:
Thanks @JaGeo!

PR #389, opened 2022-04-07
Incorrect parameters for hybrid functionals

nwinner:
PBE0 and HSE have 75% GGA exchange, not 100% (https://www.vasp.at/wiki/index.php/List_of_hybrid_functionals)

Change the name of PB0 to PBE0 (the correct name)

Change the name of ALDCAC to ALDAC (https://www.vasp.at/wiki/index.php/ALDAC)


munrojm:
Thanks @nwinner!

PR #403, opened 2022-04-26
[WIP] `TaskDoc` changes

munrojm:
This PR merges the task document models into a single one that can be used by both the validation builder, and the API client.


mkhorton:
What is meaning/possible values of `state`? Why the change in import/filename?

mkhorton:
Ah, I see `Status` -- should this not be `State`? (also typo in `SUCESS`)

munrojm:
I have renamed it.

I have had a goal to merge the two task doc models in emmet-core for a while now. I think I may just use this PR to do it... These changes were in response to a meeting with @acrutt.

mkhorton:
Makes sense, I think I was just confused by the `_valid`. May this not cause confusion because a `TaskDocument` with state `VALID` may still generated a `ValidationDoc` with `.valid == False`?

arosen93:
Looks like this can potentially be closed, @munrojm?

PR #414, opened 2022-05-11
Rename old molecules module

munrojm:
Rename the old molecules module to allow access alongside the new molecule additions.


mkhorton:
Suggest `molecules_jcesr` rather than `molecules_old`. As far as I know, new molecules are still being added to this collection (we just added a few thousand in 2022).

PR #424, opened 2022-05-25
Update the description for chemsys_permutations

acrutt:
Update the description for chemsys_permutations because the previous description was inaccurate


munrojm:
Thanks @acrutt!

PR #425, opened 2022-05-25
Add optic absorption builder 

yang-ruoxi:
For building absorption spectrum of materials on MP

## Contributor Checklist

- added AbsorptionDoc
- added absorption_spectrum.py for building the items from calculation tasks.


munrojm:
Thanks @yang-ruoxi! This looks great!! One thing I would add to the model would be default values of `...` for required fields in the document. If a field is not required, then I would set the default value to `None` like for`material_id`.

mkhorton:
Just a small comment, I would maybe be more careful with the Field descriptions: these become public and are part of the public documentation. So, for example, `is_hubbard`, "whether the material is hubbard" may not be useful to a newcomer (I leave this comment realising this may have been copied from somewhere else).

mkhorton:
Just to share for @munrojm's benefit, @yang-ruoxi and I met yesterday to outline this builder and clarify intent. This is what we came up with:

image

mkhorton:
There were a few points of confusion, the pre-chunk code I'm not sure if that can be inherited or better documented somewhere.

What we did not discuss is how or if we want to integrate this into electronic structure visualizations or apply scissor corrections to the absorption if HSE gaps are in the database: perhaps these could be topics for a future PR.

yang-ruoxi:
> There were a few points of confusion, the pre-chunk code I'm not sure if that can be inherited or better documented somewhere.
> 
> What we did not discuss is how or if we want to integrate this into electronic structure visualizations or apply scissor corrections to the absorption if HSE gaps are in the database: perhaps these could be topics for a future PR.

Agree with the HSE gaps. Would like to follow up on this for future PR

PR #426, opened 2022-05-27
Jaguar and Reactions

espottesmith:
This PR adds infrastructure (data models, builders) to store and manipulate output from Jaguar electronic structure calculations. The main purpose of this is to compile individual calculations of potential energy surface (PES) minima and transition-states (TS) to form reaction documents. This will be used for an upcoming dataset of electrolyte reactions and their properties.

Currently, there the builders have one important limitation: the builder only works if there are reaction endpoints (reactants and products) which have been successfully optimized from the TS. If the endpoint optimization failed because the reactants and products move towards infinite separation, the reaction cannot be built. This happens infrequently, so it does not significantly limit the utility of this work.

Future work will include:
- Addressing the limitation above, allowing for calculation of reaction properties where endpoints are calculated at infinite separation
- Builders to link endpoints calculated in Jaguar to their associated molecules calculated in Q-Chem and stored in the molecules collections.
- Potentially expanding the reaction properties of interest (for instance, including changes in charge states/partial charges, bonding information via e.g. NBO analysis, etc.) 
- Allowing for reactions without transition-states (especially charge transfer reactions between molecules)

All new data models and builders have tests associated with them (which pass on my local system)


espottesmith:
Might eventually reopen this or open a related PR. But as I'm the main person in our ecosystem using Jaguar for high-throughput DFT (and that relies on an ongoing partnership with Schrodinger), I'm not sure how much sense it makes to include this in the main emmet repo.

PR #440, opened 2022-06-22
Molecules update

espottesmith:
(Mostly) small changes to the recently included molecule builder pipeline. The biggest change is in creation of an MPculeID, a new ID format. Previous discussions led me to believe that an ID based on task IDs is fragile, potentially breaking if old tasks become deprecated. In order to avoid this problem while also guaranteeing the uniqueness of IDs, the MPculeID follows the format:

(prefix-)hash-charge-spin

where the prefix is optional and the hash is a Weisfeiler Lehman on a molecular graph representation (generated using networkx). The node attributes used to generate the WL hash are either the XYZ coordinates of the molecule (used for molecule association) or the atom species (used for constructing final molecule docs). Because charge and spin are also included in the ID, the risk of IDs no longer being unique due to a hash collision becomes negligible. To further protect against this, we could add the alphabetical formula, so that the ID would become (prefix-)hash-formula-charge-spin.

I note that these IDs are not easily understood or remembered by humans but can be easily generated (there is a utility function in emmet.core.utils; this utility might eventually be better placed in pymatgen).

## Contributor Checklist

- [X] I have broken down my PR scope into the following TODO tasks
   -  Make it so bonding is not a required component of a build pipeline
   -  Fix SMD string representation for G2 solvent
   - Add MPcule ID format
- [X] I have run the tests locally and they passed.
- [X] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR


munrojm:
@espottesmith are you good if I merge when the tests pass? Is there any remaining issue you would like to discuss?

espottesmith:
I'm making some changes in response to @rkingsbury's comments. I'll let you know when I'm ready to merge.

emmet-builders/emmet/builders/qchem/molecules.py
```diff
@@ -224,29 +223,38 @@ def process_item(self, items: List[Dict]) -> List[Dict]:
...
             try:
-                molecules.append(MoleculeDoc.from_tasks(group))
+                doc = MoleculeDoc.from_tasks(group)
+                molecule_id = "{}-{}-{}".format(
+                    doc.coord_hash,
+                    str(int(doc.charge)).replace("-", "m"),
+                    doc.spin_multiplicity,
+                )
+                doc.molecule_id = molecule_id
+                molecules.append(doc)
```
rkingsbury, line 242: @espottesmith when testing out the build pipeline on solvated ion clusters I found a bug in this section of code. For monatomic ions with the same charge, this will result in duplicate `molecule_id` b/c the coordinates, charge, and multiplicity can all be identical.

I worked around the issue by doing this; although there is probably a more robust / general solution:
```
        for group in self.filter_and_group_tasks(tasks):
            try:
                # TODO - this molecule_id formula will return identical ids for single atoms
                # that have the same charge and spin
                doc = MoleculeDoc.from_tasks(group)
                # use something different for single atoms
                if len(doc.composition)==1:
                    molecule_id = MPculeID("{}-{}-{}".format(
                        doc.species_hash,
                        str(int(doc.charge)).replace("-", "m"),
                        doc.spin_multiplicity,
                    ))
                else:
                    molecule_id = MPculeID("{}-{}-{}".format(
                        doc.coord_hash,
                        str(int(doc.charge)).replace("-", "m"),
                        doc.spin_multiplicity,
                    ))
                doc.molecule_id = molecule_id
                molecules.append(doc)
```

I tried just changing the `coord_hash` to `species_hash` for all tasks, but that can result in duplicated IDs for other reasons, so I think it's important to keep the coordinates in there in most cases.

espottesmith: Yeah, I realized this a while back and hadn't had a chance to fix it. The solution that I'm going to go for is actually just making MPculeID a little longer and including the formula. I actually also think that'll help it be more readable. The hash still won't be easily interpreted, but the user will at least have a clue what they're looking at if the formula is there.

rkingsbury: Sounds good. If you have any prototype code for that (even if very rough), feel free to send my way and I can test it on my builds (but if you haven't got there yet, no worries)

rkingsbury: I just realized we also need to consider what happens if we have the same single atom / ion in different solvents. My solutions breaks in that case - all solvents collapse to the same MP-ID which results in calcs getting lost in the build pipeline.

espottesmith: That should be fine, I think? Since I've changed how level of theory is handled, and molecules in different solvents should be collapsed into a single document anyways. You're right that this will be treated somewhat differently in the association stage, because in general molecules in different solvents won't have identical structures. But I don't believe it's inconsistent. Will test this to confirm.

rkingsbury: Yeah as long as molecules in different solvents wind up in the same document / ID this should be OK. I'm not seeing that happen in my current build pipeline testing (it seems that only Water gets past the association stage), but I haven't had time to troubleshoot much yet

espottesmith: If you can send me a minimal test case (let's say one molecule in 2 or 3 different solvents), that'd actually be a big help.

espottesmith:
In response to a series of conversations with @rkingsbury, @samblau, @arosen93, and Orion Cohen, I'm making a bunch of changes to how the molecules builders work. The main purpose is to better handle cases where the properties of a molecule are calculated in multiple solvent environments. I'll leave a comment when these changes are made and are stable. In the mean time, I'll probably be breaking a lot of things and annoying lots of folks with notifications.

munrojm:
@espottesmith no problem on the notifications. That all sounds good!

espottesmith:
This PR is ready for review - @rkingsbury @munrojm.

Also, what's up with testing? Seems that on the main branch, many test files had been deleted?

espottesmith:
Found an edge case where monatomic hydrogen (H1) was causing certain builders to fail. Resolved now - have built collections involving 48,672 unique molecules.

munrojm:
> Found an edge case where monatomic hydrogen (H1) was causing certain builders to fail. Resolved now - have built collections involving 48,672 unique molecules.
> 
> Should be good to merge now.

Great, that sounds good. Working to get the tests fixed, then I will merge.

rkingsbury:
@espottesmith apologies I haven't had a chance to do much testing with this yet. It is still on the docket though, possibly later this week / early next. I don't have any objections to merging, but if there's anything in particular you want me to test out please let me know. I'd say this is a tremendous foundation for molecule docs though, and we can always address small issues in subsequent PRs.

PR #437, opened 2022-06-16
Migrationgraphdoc

hmlli:
*Start with a description of this PR. Then edit the list below to the items that make sense for your PR scope, and check off the boxes as you go!*

## Contributor Checklist

- [X] I have broken down my PR scope into the following TODO tasks
   -  [X] add file for migrationgraphdoc
   -  [x] all functionalities added to migrationgraphdoc
- [x] I have run the tests locally and they passed.
- [x] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR

PR #468, opened 2022-07-21
Add Migration Graph Builder

acrutt:
Adding a new MapBuilder to create migration graphs from an InsertionElectrode collection. Based on MigrationGraphDoc developed by @hmlli in https://github.com/materialsproject/emmet/pull/437


munrojm:
Thanks @acrutt  and @hmlli !! 

I remember @hmlli originally having a try-except block in a document construction method. If document construction failure is possible, but the desire is to exclude the failure and keep processing, then a try-except block should be added to the builder processing method.

acrutt:
@munrojm I finished updating the builder after @hmlli latest changes to the MigrationGraphDoc. I think this is ready to merge unless you have any suggested edits!

PR #482, opened 2022-07-27
Update Redox Elements

acrutt:
Per the June 13, 2022 meeting and discussion with Kristin and other Persson Group members, we would like to make the InsertionElectrodes builder consistent with the screening process used for identifying intercalation electrodes in research. Therefore the list of redox active elements was updated to remove Hf (only affects two battery entries on MP) and add Ag.


munrojm:
Looks good!

PR #483, opened 2022-07-27
Permitting Electrodes with Extractable Ions

acrutt:
The previous StructureGroupBuilder base query did not permit the processing of structures that contained ["Li", "Be", "Na", "Mg", "K", "Ca", "Rb", "Sr", "Cs", "Ba"]. However, it is possible to have insertion electrodes that contain extractable ions (e.g. ["H", "Li", "Na", "K", "Rb", "Cs", "Ag", "Cu"]) and these types of systems were being excluded from the InsertionElectrodeBuilder.

This requirement has been removed from the StructureGroupBuilder to allow the analysis of Insertion Electrodes with extractable ions. Note this is a significant change because the Insertion Electrodes will no longer be limited to systems that only contain a single mobile species, however all analysis will still only assume one species as the working ion.

### DIFF
--- a/emmet-builders/emmet/builders/materials/electrodes.py
+++ b/emmet-builders/emmet/builders/materials/electrodes.py
@@ -154,14 +154,12 @@ class StructureGroupBuilder(Builder):
             - get the oldest timestamp for the target documents (min_target_time)
             - if min_target_time is < max_mat_time then nuke all the target documents
         """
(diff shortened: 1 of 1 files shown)


munrojm:
@acrutt Since the builders are currently using `["Ca", "Cs", "Rb", "K", "Y", "Na", "Al", "Mg", "Li", "Zn"]` as the list of working ions, we should probably settle on the final list before I merge this.

PR #491, opened 2022-08-01
[WIP] Updated StructureGroupDoc

kim-jiyoon:
I added two fields to the StructureGroupDoc so that the document returns the identified "host" material id in a given structure group as well as the remaining "insertion" material ids (non-host material ids). The "host" is corresponds to the structure with the lowest concentration of "working ion"/"ignored specie" in a given structure group, and if there are multiple that share the lowest concentration, the one with the lowest energy is considered as the "host". I also updated the StructureGroupDoc to specify 1 "ignored specie" instead of multiple, because the StructureGroupDoc is only called in the StructureGroupBuilder, which is used to consider 1 "ignored specie" at a time when structure grouping.

@munrojm @acrutt


emmet-core/emmet/core/structure_group.py
```diff
@@ -193,11 +214,29 @@ def from_ungrouped_structure_entries(
...
+    ) -> dict:
+        host_and_insertion_ids = {"host_id": None, "host_entries": [], "insertion_ids": []}  # type: dict
+        ignored_specie_min_fraction = min([e.composition.get_atomic_fraction(ignored_specie) for e in entries])
+
+        for e in entries:
+            if e.composition.get_atomic_fraction(ignored_specie) == ignored_specie_min_fraction:
+                host_and_insertion_ids["host_entries"].append(e)
+            else:
+                host_and_insertion_ids["insertion_ids"].append(e.entry_id)
+        host_and_insertion_ids["host_id"] = min(host_and_insertion_ids["host_entries"],
```
acrutt, line 230: If I'm understanding this correctly, it looks like among the host entries, the one which is selected for the host_id is the oldest entry? Do we want less arbitrary criteria, like selecting the host entry with the lowest energy instead?

acrutt: Misread the code. Please disregard this comment!

munrojm:
Thanks for the effort here @kim-jiyoon! Merging now.

PR #494, opened 2022-08-05
StructureGroupBuilder and StructureGroupDoc minor fixes

kim-jiyoon:
There are 2 small bugs found and things should now be working as expected after testing the StructureGroupBuilder on the MP materials collection and not just my personal materials collection.

1) The field name was not updated for the "host_material_ids" since a small change was made to the field
2) The StructureGroupBuilder and StructureGroupDoc originally did not consider energy data, but now that it is being used to determine the lowest energy "host_material_id" when there are multiple potential ids to choose from, the existing method of generating a ComputedStructureEntry in the " _entry_from_mat_doc" method needed to be revised. This is a more robust method because it uses all the information available in the MaterialsDoc's entries field and checks for the run type, which is not always just GGA or GGA+U especially in the MP materials collection. The MP compatibility scheme can only apply to calculations performed with GGA or GGA+U, and GGA+U is prioritized because it is a higher level of theory.

@munrojm  @acrutt


munrojm:
Hi @kim-jiyoon, this looks good to me! Thank you for the fixes.

PR #497, opened 2022-08-08
Aneb utils func

hmlli:
A simple utils function that takes a `MigrationGraphDoc` object and converts it into a set of aneb wf inputs. This is to avoid unnecessary redundant end point calculations, as not all sites in the doc are used in the ANEB calculation.

### DIFF
--- a/emmet-core/emmet/core/mobility/utils.py
+++ b/emmet-core/emmet/core/mobility/utils.py
@@ -1,0 +1,6 @@
+from emmet.core.mobility.migrationgraph import MigrationGraphDoc
+
+def get_aneb_wf_inputs(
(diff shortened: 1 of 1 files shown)


hmlli:
@munrojm  Hi Jason! I'm adding a utils function to help convert the `MigrationGraphDoc` to aneb workflow inputs on `atomate`. Compatibility for `atomate2` is on the table when the `atomate2` version of aneb wf is done (I plan to work on this some time in the future.)
That being said, I've added a `utils.py` under `emmet.core.mobility`. Is this appropriate? Please let me know if you have any questions, whether I should move the function to somewhere else, or whether this utils function should be in this code base at all. Thanks!!

PR #508, opened 2022-08-24
Add chunk size to thermo

munrojm:
Adds chunk size as an explicit input parameter to thermo to ensure proper serialization.


acrutt:
@munrojm I think you may also need to add `chunk_size=chunk_size, ` to line 96 in order to make sure the set chunk_size gets passed through. Otherwise the chunk_size is added as an attribute, but remains set at the default value even if a custom chunk_size is set.

PR #510, opened 2022-08-25
InsertionElectrodeDoc material_id fix

acrutt:
InsertionElectrodeDoc replace entry.entry_id with entry.data["material_id"] to correctly pull material id after entry changes


acrutt:
@munrojm I think this should be ready to merge!

PR #514, opened 2022-08-26
Migration Graph Builder bug fix

acrutt:
typo in migration graph builder caused .as_dict() to fail. Fixed typo and added test.


acrutt:
@munrojm I think this should be ready to merge too!

PR #520, opened 2022-08-30
Remove get_summary_dict() from InsertionElectrodeDoc

acrutt:
The `InsertionElectrodeDoc.from_entries()` relied on pymatgen's `InsertionElectrode.get_summary_dict()`. With the most recent changes to the thermo builder (where material_ids are no longer stored as entry.entry_id), this method was no longer reliable for correctly extracting and storing material_ids. `InsertionElectrodeDoc.get_elec_doc()` was added to replace `get_summary_dict()` to resolve this bug. Overall this change will also allow other document features to be customized based on Material Project specific needs moving forward (independent of pymatgen).

Minor clean-up to `InsertionElectrodeDoc.from_entries()` was added to simplify this code.


acrutt:
@munrojm I think this is ready to merge! Let me know if any edits are needed.

PR #522, opened 2022-09-01
Make robocrys and pmg addon packages optional

munrojm:
- `robocrys`, `pymatgen-analysis-diffusion`, and `pymagen-analysis-alloys` are all now optional in `emmet-core`

### DIFF
--- a/emmet-api/setup.py
+++ b/emmet-api/setup.py
@@ -21,7 +21,7 @@ setup(
     url="https://github.com/materialsproject/emmet",
     packages=find_namespace_packages(include=["emmet.*"]),
     install_requires=[
-        "emmet-core",
+        "emmet-core[all]",
         "fastapi",
         "uvicorn",
         "gunicorn",
--- a/emmet-builders/setup.py
+++ b/emmet-builders/setup.py
@@ -20,7 +20,7 @@ setup(
     author_email="feedback@materialsproject.org",
     url="https://github.com/materialsproject/emmet",
     packages=find_namespace_packages(include=["emmet.*"]),
-    install_requires=["emmet-core", "maggma>=0.47.3", "matminer>=0.7.3"],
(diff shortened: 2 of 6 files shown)


mkhorton:
How are you making packages optional when the document model depends on them, e.g. I believe `pymatgen-analysis-alloys` has `AlloyPair` in the document model definition?

munrojm:
Yes, it does. I've just added a check to the beginning of the module import statements. That way if you go to import the model class it tells you to install the package if you don't have it. They modules are self-contained in emmet-core so I figured this solution would be okay. I was under the assumption we absolutely don't want pymatgen to depend explicitly on the addon packages.

munrojm:
Happy to do something different if you feel there is a better solution.

mkhorton:
> I was under the assumption we absolutely don't want pymatgen to depend explicitly on pymatgen-analysis-diffusion or pymatgen-analysis-alloys.

That's fair; maybe. I think the main thing is to ensure that the API is still usable for people who don't have those installed, e.g. that we ensure that the error messages if you're trying to use the `.alloys` (or, similarly, any routes that depend on diffusion, e.g. wherever the migration hop API is going to live) get a nice error message which says:

> To use the _____ API, please install the additional package ______ with a command like "pip install _______".

I think the error message on the import in emmet-core is ok, but wouldn't be sufficient for the client users, so perhaps we can just add another `try/except ImportError` in the client code.

munrojm:
Yeah, my plan was to do just that and add another try-except block in the client.

mkhorton:
I think this makes sense! I can vouch for `pymatgen-analysis-alloys` as being fairly light on the dependencies, but nevertheless it does add one (`shapely`) so I think this strategy is better.

Maybe we can slightly modify the API installation instructions to be `pip install mp-api[all]` or something for people who install the client directly.

munrojm:
> Maybe we can slightly modify the API installation instructions to be pip install mp-api[all] or something for people who install the client directly.

Sure, that sounds good to me!

PR #529, opened 2022-09-08
Remove debugging print statement

acrutt:
Remove debugging print statement from Migration Graph builder


acrutt:
@munrojm the segmentation fault error is popping up again? But it looks like the testing environment is using an older version of spg. 

Collecting spglib>=1.9.9.44
  Downloading spglib-2.0.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (514 kB)

munrojm:
I was under the impression it was an issue with `spglib >= 2.0` which I have fixed in emmet-core

acrutt:
@munrojm tests are passing now so I think we should be good!

munrojm:
Thanks @acrutt!

PR #562, opened 2022-10-20
Fix thermo builder and phase diagram generation

munrojm:
- `ThermoBuilder` now uses `PatchPhaseDiagram` to ensure all phase diagrams are built and stored
- `SummaryBuilder` now works with new `ThermoDoc` model


jan-janssen:
Is there a specific reason why `spglib` is restricted to `<2.0.0`? Basically reverting https://github.com/materialsproject/emmet/pull/555 ?

PR #565, opened 2022-10-21
Fix pip install from source

jan-janssen:
Example how the installation from source fails: 
https://github.com/jan-janssen/emmet-test


munrojm:
Okay, this looks good. Thank you for catching this and proposing a fix. Once the tests pass with my small additions I will merge and patch release!

jan-janssen:
Thanks a lot

PR #566, opened 2022-10-22
Fix use of `PatchedPhaseDiagram` in `ThermoBuilder`

munrojm:
Ensures base phase diagram always gets built when using `PatchedPhaseDiagram` in `ThermoBuilder`.


emmet-api/setup.py
```diff
@@ -8,9 +8,9 @@
 setup(
     name="emmet-api",
     use_scm_version={
-        "root": ".",
+        "root": "..",
```
jan-janssen, line 11: This basically reverts https://github.com/materialsproject/emmet/pull/565

munrojm: Yes, I had to revert as the git information wasn't properly being accessed by setuptools when root is set to `.`

I will try and find another solution, but in the mean time this change needed to happen so our deployment pipelines don't break.

Issue #595, opened 2022-11-23
ChemEnv integration


JaGeo:
As mentioned by mail, I would like to provide the ChemEnv data, @munrojm.

I am currently testing several options:
- full run based on current implementation (12 hours on 24 cores for me, our system has a lot of memory available)
- coordination environments only for structures which have oxidation states  (5 hours on 24 cores)
- coordination environments only for structures which have less than or equal to 100 inequivalent sites in the structure to reduce the run times 
- combination of both options

I will add run times and provide potential implementations in the form of a pull request in a next step. We should potentially test the fastest option first and then see if this is doable on your systems and potentially extend the number of compounds complexity step by step.

JaGeo:
I think I will just make a pull request for the option where we only do the analysis when oxidation states are available. I will uncomment the other code for now. It runs at least twice as fast and should also be easier to understand for users if only one type of ChemEnv analysis is used.

munrojm:
@JaGeo, just pinging you here to let you know I am debugging some builder pipeline issues which is delaying me on getting the `chemenv` stage up and running. This should be done soon, and I will update this thread.

JaGeo:
@munrojm Thanks for the update! 🙂

munrojm:
@JaGeo, the builder runs well! I am able to get it completed in ~5hrs within our build pipelines. The data should be incorporated into the DB and available through the API next release.

JaGeo:
@munrojm  Awesome! 🥳 Thank you so much!

JaGeo:
@munrojm If there is more documentation needed to make this available, just let me know. I am happy to help with this as well.

munrojm:
@JaGeo, I appreciate the offer. Apologies again for the delay on this. This is essentially ready to go on the API side. Our initial intent was to wait to bundle this with some other data in the next DB release, but that kept getting pushed back for a number of reasons. Currently, I am working to fix up thermo data related issues. We intend to deploy again soon with to accommodate the changes. What I will do is include the ChemEnv data in that deployment so it is available next week.

JaGeo:
Great, thank you!

JaGeo:
I do have an additional question: would it be in principle possible to get a DOI for such a dataset (the whole one)?

munrojm:
[...]

PR #596, opened 2022-11-23
ChemEnv only when oxidation states exist

JaGeo:
As mentioned in https://github.com/materialsproject/emmet/issues/595, this implementation reduces the run times by more than 50% and should therefore  hopefully allow to integate the ChemEnv data


JaGeo:
@munrojm Tests are passing. I am happy to hear what you think and I am also interested whether the builder now works on your systems and with your memory requirements etc.

munrojm:
@JaGeo, thank you for these changes! This looks great. I am going to merge it now, and will get it running in our build pipeline tomorrow. I will update https://github.com/materialsproject/emmet/issues/595 once it is is chugging along with any comments or concerns.

PR #673, opened 2023-03-08
Utils to static

hmlli:
turn the utils function in MigrationGraphDoc into a static method, so that the MigrationGraphDoc object doesnt have to be initiated to use this method. Cleaner for use case of pulling data from db.


hmlli:
@munrojm Hi Jason! just a small change in a utils function within `MigrationGraphDoc`. Basically in the usage case, the function is almost never used with the instance initiated, so I changed it to a static method. Thanks!

munrojm:
Sounds good! Thanks @hmlli.

PR #693, opened 2023-04-09
[WIP] Port the cclib TaskDocument to emmet

arosen93:
Context: There is a package called [cclib](https://github.com/cclib/cclib) that serves as a community-developed set of parsers for molecular DFT codes with a standardized schema. In Atomate2, I developed a `TaskDocument` centered around cclib to make it easy to generate `TaskDocument` objects for a wide range of codes. This _greatly_ reduces the time it takes to develop workflows for new codes and does so in an internally consistent way. We didn't add `cclib` as a required dependency; it was optional, but if the user had it installed, they could have a lot of flexibility with this `TaskDocument`.

In hindsight, I think this is better ported to `emmet` since it is quite general. It is a pretty easy move, and that's what this PR tries to achieve. Links to [here](https://github.com/materialsproject/atomate2/issues/302) in the Atomate2 repo Issue tracker. I use this `TaskDocument` heavily in my [QuAcc](https://github.com/arosen93/quacc) code, so it will be of immediate use. QuAcc is currently dependent on Atomate2 solely for this class.

Happy to get your thoughts on if this is the right home for it.


arosen93:
Alright, it looks like this is ready to go. @utf, could I get your approval on moving the cclib `TaskDocument` before Jason signs off on this? Thanks!

PR #705, opened 2023-04-27
Task doc validation changes

munrojm:
This PR make some small changes to how some of the `TaskDoc` fields are validated to ensure it can be used with the `mp-api` client.

### DIFF
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -334,7 +334,7 @@ class TaskDoc(StructureMetadata):
     )
     additional_json: Dict[str, Any] = Field(None, description="Additional json loaded from the calculation directory")
 
-    custodian: List[CustodianDoc] = Field(
+    custodian: List[Dict[str, Any]] = Field(
         None,
         title="Calcs reversed data",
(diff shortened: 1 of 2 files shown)


munrojm:
@utf, do you have any issues with these small changes? I've tested this with the current client and a large amount of the task collection data.

munrojm:
I'm realizing the changes are hard to see because of the formatting. Here is a summary:

- custodian `corrections` set to `List[Any]`
- `task_type` validated as both `CalcType` or `TaskType`
- `CalculationInput` now also uses pmg `Kpoints` object in validation
- `epsilon_*` matrices swapped to `Tensor3R` for `Tuple` -> `List` validation for `[]` cases

utf:
Hi @munrojm, those changes all look fine to me!

utf:
Hi @munrojm, this is causing issues with atomate2 task validation.

The issue is that the epsilon_* matrices are not rank 3 tensors, they are rank 2 tensors.

Is there a reason you changed them to rank 3?

munrojm:
@utf, that is definitely a mistake and should be switched. The reason I switched the type was related to the Tuple vs List validation issue for cases when it is empty.

PR #716, opened 2023-05-08
correct import of EmmetSetting

marcodigennaro:
Simple PR to correct import of EmmetSetting in documentation

The previous import 
```from emmet.settings import EmmetSettings```
would lead to 
```ModuleNotFoundError: No module named 'emmet.settings'```


munrojm:
Thanks @marcodigennaro!

PR #719, opened 2023-05-08
Change type of epsilon_*

utf:
Change epsilon_* to be a rank 2 matrix as a list as discussed in #705


munrojm:
No worries. As soon as the tests are done I will merge and patch release.

PR #326, opened 2022-01-04
 Elasticity doc and builder

mjwen:
Schema and builder for elastic tensors and derived properties. 

This PR regenerates the builders for elastic tensor calculations. It

- creates schema for elastic tensors, compliance tensors, derived properties, as well as fitting data; the hope is to enable full provenance tracking and regeneration of the docs. 

- ports three different elastic tensor fitting methods from atomate, which can be selected at the builder level

- filter tasks by INCAR params and task completion time in causes multiple tasks are associated with the same material 

TODO: more unit tests should be added

PR #721, opened 2023-05-11
Elastic builder 

mjwen:
This PR further improves the elastic builder by symmetrizing the stresses before doing the fitting. This is needed because when `sym_reduce` is turned on in the workflow, the stresses of symops-related structured can be slightly off. But this may result in large difference in elastic tensors in the fitting. Symmetrizing the stresses resolve the problem. 

I believe the commits before Dec 15, 2022  have been previously merged in #326. Not sure why it shows up again.


mjwen:
hmmm, seems a lot of conflicts due to formatting. I guess I was using black default (i.e. 88). From [.pre-commit-config.yaml], black is used as well, but the line length is not the default. What are we using for the repo?

munrojm:
@mjwen the easiest thing to do would be to install pre-commit locally and run `pre-commit run --all-files` in the root dir of the repo and then commit all of the change it makes.

mjwen:
@munrojm I have run it locally. The upstream code seems hasn't formatted according to the pre-commit file. Can you have a quick run, at least for 

`emmet-builders/emmet/builders/materials/elasticity.py` and 
`emmet-core/emmet/core/elasticity.py`

in the main branch? 

After that, it would be much easier for this merge.

munrojm:
Oh okay that is probably my fault then, yes let me do that so this can be merged more easily.

munrojm:
@mjwen, okay you should be good to rerun the pre-commit stuff and merge from main.

mkhorton:
Should any of these changes be incorporated into the atomate2 workflow? Unclear to me if there’s an opportunity to unify elastic document models here. Tagging @utf for visibility.

mjwen:
@mkhorton  I think we should... Just started a similar discussion in another stomate2 issue https://github.com/materialsproject/atomate2/issues/333

mjwen:
I've rebased this on materialsproject:main, but it cannot pass the linting.

The problem is that we are using inconsistent linting settings in [.pre-commit-config.yaml](https://github.com/materialsproject/emmet/blob/main/.pre-commit-config.yaml) and GH actions https://github.com/materialsproject/emmet/blob/dd20f1fba949a8754d4c4578da808a3b9d83abd3/.github/workflows/testing.yml#L32

I guess we want to update the GH actions to use .pre-commit-config.yaml? Replacing the flake8 stuff by https://github.com/materialsproject/atomate2/blob/b4a87a38a8b02a0b73284ce029acf9d97ea1b6a6/.github/workflows/testing.yml#L21 should do the job.

munrojm:
Yes, I have been in the process of overhauling the workflows. Will switch the linting over fully now so it works here.

munrojm:
@mjwen should be fixed.

mjwen:
@munrojm thanks! 

A last bit: do you want to move [mypy stuff](https://github.com/materialsproject/emmet/blob/435a19abfd6915385cdc837dde76fcadd42e8046/.github/workflows/testing.yml#L97) to recommit? 

If there are concerns, I can run mypy locally, and modify my PR. Either way is fine.

mjwen:
Ready for review.

munrojm:
I might remove mypy actually, so I can deal with that. I appreciate the offer, though.

This looks good to me! Thank you for all the work put into these changes.

Issue #735, opened 2023-05-20
Merging atomate2 calc_types schema 


mjwen:
Atomate2 is using emmet schema now, but the `calc_types` schema still exists in both repo: [emmet calc_types](https://github.com/materialsproject/emmet/tree/main/emmet-core/emmet/core/vasp/calc_types) and [atomate2 calc_types](https://github.com/materialsproject/atomate2/tree/main/src/atomate2/vasp/schemas/calc_types). 

There are still some stuff not moved from atomate2 calc_types to emmet. For example, the [MD](https://github.com/materialsproject/atomate2/blob/049c7412cc0fc6948b570808130f932f9477fcfa/src/atomate2/vasp/schemas/calc_types/_generate.py#L19) option is not in emmet (see [here](https://github.com/materialsproject/emmet/blob/47ae22c6071319254fac1c0dd8d6b8ebed57717b/emmet-core/emmet/core/vasp/calc_types/generate.py#L21)). As a result, atomate2 MD calculations will be assigned a calc_type `Unrecognized` now. 

I guess we want to fully merge these, and then delete the corresponding codes in atomate2, right? Any concerns? 

@utf you should have a better idea.

utf:
Thanks for raising this.  We should definitely merge the atomate2 and emmet calc_types. Would you be able to look into that?

tsmathis:
Closing since this was addressed in #764

PR #764, opened 2023-06-27
Combine atomate2 vasp calc types

mjwen:
This PR addresses #735. The below has been performed:
- add `Molecular Dynamics` task type 
- regenerate enums.py

TODO:
- [x] check the update in enums.py is OK, e.g. `PB0->PBE0` and removing `PBEsol` (now using `PBESol`?).  
- [x] use it from the atomate2 side to make sure everything works well 
- [x] figure out what the line `elif len([x for x in kpts.get("labels") or [] if x is not None]) > 0:` in `utils.py` is doing and possibly rewrite it. (@utf can give a quick answer).


mjwen:
- For the 1st TODO, after I rerun `generate.py`, these changes happen. I guess they are supposed to happen and it is just we haven't run `generate.py` after modifying the other files? 

- For the 3rd TODO, [the line](https://github.com/materialsproject/emmet/blob/8e84337854488809934953eaefc6fd6d641a3699/emmet-core/emmet/core/vasp/calc_types/utils.py#L81) is copied `atomate2`. I haven't no idea whether we need to add this line or not. If yes, can we simplify it? 

Anyone knows these more please comment and we can then move forward.

utf:
> figure out what the line elif len([x for x in kpts.get("labels") or [] if x is not None]) > 0: in utils.py is doing and possibly rewrite it.

I can't remember what this is doing. But I would be reluctant to remove it to avoid any breaking changes.

mjwen:
Ready for review!

utf:
Just checking if there has been any progress on this? Ideally we can get this merged to progress to a new atomate2 release.

munrojm:
Sorry for the delay, I will take a look today and test it myself today.

utf:
Thanks @munrojm!

emmet-core/emmet/core/vasp/calc_types/enums.py
```diff
@@ -1,18 +1,18 @@
...
 
 class RunType(ValueEnum):
-    """VASP calculation run types"""
+    """VASP calculation run types."""
 
     AM05 = "AM05"
     GGA = "GGA"
     PBE = "PBE"
     PBESol = "PBESol"
-    PBEsol = "PBEsol"
```
matthewkuner, line 15: @mjwen Why was this changed from PBEsol to PBESol? This (very minorly) breaks some of the validation code I'm working on. Also, it isn't consistent with the original PBEsol paper's capitalization (https://arxiv.org/abs/0711.0156)

mjwen:
Sorry that this causes trouble for you! 
 
The change was introduced when merging this and the atomate2 enums.py, where [PBESol](https://github.com/materialsproject/atomate2/blob/4e46b73afcac6ba3b35cc40a46b0e160fd7aed36/src/atomate2/vasp/schemas/calc_types/enums.py#L15). 

I am OK to change it back, but we need to ensure this will not introduce any problem for atomate2. 

@utf Any idea on this? Will atomate2 get into trouble if we use `PEBsol`? Or do you prefer `PBESol`?

mjwen:
Related, we also did `PB0->PBE0`. Not sure if this will cause trouble for others or not...  @munrojm any thoughts?

PR #765, opened 2023-06-28
Cli updates

tsmathis:
This PR addresses two bugs in the CLI that occur when trying to parse calculations into an empty task collection.

1. (utils.py) Starting with Pymongo4(Maggma currently requires pymongo>=4.2), the collection.ensure_index() function has been replaced with collection.create_index() [pymongo4 migration, ensure_index](https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#collection-ensure-index-is-removed). This function is only called when trying to parse task documents into a collection that does not already have the required indexes.
2. (tasks.py) Trying to reserve `task_id`s in an empty collection results in an `IndexError: list index out of range` error. Specifying a `task_id` of 1000001 fixes this. The 7 digit default `task_id` also ensures any task documents in the test collection will not be overwritten

### DIFF
--- a/emmet-cli/emmet/cli/tasks.py
+++ b/emmet-cli/emmet/cli/tasks.py
@@ -611,7 +611,12 @@ def parse(task_ids, snl_metas, nproc, store_volumetric_data, runs):  # noqa: C90
             {"$group": {"_id": None, "num_max": {"$max": "$num_int"}}},
         ]
         result = list(target.collection.aggregate(pipeline))
-        next_tid = result[0]["num_max"] + 1
+
+        if not result:
+            next_tid = 1000001 # Only occurs when inserting into an empty task collection for testing
+        else:
+            next_tid = result[0]["num_max"] + 1
+
         lst = [f"mp-{next_tid + n}" for n in range(nmax)]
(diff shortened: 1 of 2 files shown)


emmet-cli/emmet/cli/tasks.py
```diff
@@ -611,7 +611,12 @@ def parse(task_ids, snl_metas, nproc, store_volumetric_data, runs):  # noqa: C90
             {"$group": {"_id": None, "num_max": {"$max": "$num_int"}}},
         ]
         result = list(target.collection.aggregate(pipeline))
-        next_tid = result[0]["num_max"] + 1
+
+        if not result:
+            next_tid = 1000001 # Only occurs when inserting into an empty task collection for testing
+        else:
+            next_tid = result[0]["num_max"] + 1
+
```
tschaume, line 619:
```python
next_tid = result[0]["num_max"] + 1 if result else 1000001
```

tschaume (changes requested):
Also [fix the linting errors](https://github.com/materialsproject/emmet/actions/runs/5406888231/jobs/9824264114#step:5:55)

tschaume (approved):

PR #676, opened 2023-03-15
Merge atomate2 taskdoc

utf:
I've merged in the atomate2 VASP task document into emmet.

There are currently no breaking changes with the documents that were already in emmet. I've tried to ensure there are no changes to the task documents themselves, e.g., what is actually stored in the database rather than the schema itself. However, I would really appreciate it if you could try running any relevant builders using this branch to make sure everything still runs ok.

Here is a summary of the new fields added in this PR

- Potcar
  - pot_type
- InputDoc
  - parameters
  - psuedo_potentials
  - potcar_spec
  - xc_override
  - is_lasph
  - is_hubbard
  - hubbards
- OutputDoc
  - energy_per_atom
  - bandgap
- TaskDoc
  - Now subclasses StructureMetadata
  - structure
  - included_objects
  - vasp_objects
  - entry
  - author
  - icsd_id
  - transformations
  - additional_json
  - dir_name
- TaskType now imported from emmet.core.vasp.enums (existed already)
- CalcsReversedDoc -> Calculation from emmet.core.vasp.calculation (added in this PR)
  - Existing fields remain the same, lots of new fields added.

PR #785, opened 2023-07-18
Small fixes for `TaskDoc`

mkhorton:
Hi friends, marking this as draft so not accidentally merged. This contains some small `TaskDoc` fixes for backwards compatibility:

* `density` is now set automatically, which is important when importing older atomate2 TaskDocument (i.e. TaskDocument generated prior to #676)
* `entry.entry_id` is now set from `task_id` (and `job_id` kwarg renamed to `task_id` for consistency)
* Adds a `structure_entry` convenience property
* Ensures that `entry.parameters["potcar_spec"]` is a dict, as expected by pymatgen. 

Probably ok to merge as-is, but I'm going to look around for other issues first.

I did have some questions:

1. The `potcar_ok` validator has an `if isinstance(v, list): return [i for i in v]` line, wasn't obvious why, and didn't have a comment?
2. I'd quite like to remove/replace `icsd_id` with something more general, like `external_database_ids: Dict[str, str]` (e.g. `external_database_ids={"icsd": 1234}`, but wasn't sure if this would break anything (@utf, any comment? I think this is your addition)
3. Were there any remaining issues why `emmet.core.vasp.task_valid.TaskDocument` could not be superseded by `emmet.core.tasks.TaskDoc`?

### DIFF
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -64,12 +64,10 @@ class OrigInputs(BaseModel):
         description="Pymatgen object representing the POTCAR file.",
     )
 
-    # Make sure that the datetime field is properly formatted
     @validator("potcar", pre=True)
     def potcar_ok(cls, v):
         if isinstance(v, list):
             return [i for i in v]
-
         return v
 
     class Config:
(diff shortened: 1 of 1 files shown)


munrojm:
1. I believe the this was put in to accommodate a few of the different ways that potcar data is stored in some of the older task documents. The details are not with me, though.
3. Similarly, I don't remember the details on remaining issues here. If there are any, I don't think they are significant.

mkhorton:
Ok, in that case I think this is ready to merge as-is, unless there are issues. Any follow-up I can do in a separate PR.

Re. the potcar question, it just seemed strange to me because it looked like a no-op? Just a list -> list. Seems harmless though, and wouldn't remove it unless I understand why it was there.

munrojm (approved):
LGTM

mkhorton:
Thanks!

PR #794, opened 2023-07-24
Several `TaskDoc` improvements

mkhorton:
I noticed this was currently missing, so an important add for parity with the `task_valid.TaskDocument`.

If I have a minute, I may append to this PR to formalize the `Entry` model to ensure these fields are explicitly defined (since they're required for the compatibility scheme), otherwise the PR could be merged as-is.

### DIFF
--- a/emmet-builders/emmet/builders/materials/corrected_entries.py
+++ b/emmet-builders/emmet/builders/materials/corrected_entries.py
@@ -49,10 +49,10 @@ class CorrectedEntriesBuilder(Builder):
         self.chunk_size = chunk_size
         self._entries_cache: Dict[str, List[ComputedStructureEntry]] = defaultdict(list)
 
-        if self.corrected_entries.key != "chemsys":
+        if self.corrected_entries.key != "thermo_id":
             warnings.warn(
                 "Key for the corrected_entries store is incorrect and has been changed "
-                f"from {self.corrected_entries.key} to chemsys!"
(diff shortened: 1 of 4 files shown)


mkhorton:
I have also added a `TaskDoc.from_vasprun` method. This will log a warning, since some information from the OUTCAR will be missing, but the `TaskDoc` will still be valid and complete.

mkhorton:
Hi @munrojm, could you confirm the correct key for corrected entries? This warning seems inconsistent: https://github.com/materialsproject/emmet/blob/4ba058a2631b9fb11a8022434eb52d646de061e6/emmet-builders/emmet/builders/materials/corrected_entries.py#L52-L57

In the ThermoBuilder, it seems to suggest it's `chemsys`?

https://github.com/materialsproject/emmet/blob/4ba058a2631b9fb11a8022434eb52d646de061e6/emmet-builders/emmet/builders/materials/thermo.py#L59-L64

I can correct in this PR, but wanted to confirm first.

munrojm:
The correct key for corrected entries is indeed `chemsys`. That first block has a typo. The data for every `thermo_id` is contained in the `entries` dict, with `thermo_id` values as keys.

mkhorton:
@munrojm I think this is ready for review/merge

munrojm:
@mkhorton no concerns on my end! Looks good to me. Will merge now.

PR #811, opened 2023-08-18
[WIP] PR: Construction of a QChem TaskDoc which can eventually replace the atomate(1) drones.py functionality

rdguha1995:
*This is very much in line with Alex's vision of constructing TaskDoc's for VASP using separate models for Input and Output*

## Contributor Checklist

- [ ] There are basically two large files I am working with here
   -  [x] `calculation.py` in emmet/core/qchem has the `Calculation` class which aims to construct a Calculation doc from QChem inputs and outputs
   -  [x] `qc_tasks.py` in emmet/core/ constructs the final TaskDocument from the input and output files in a QChem calculation directory
- [x] I still need to devise some tests and I would like some help on that
- [ ] Any overall suggestions or feedback would be greatly appreciated


emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+    molecule: Molecule = Field(
+        None,
+        title="Output Molecule",
+        description="Output molecule after the QChem Calculation",
+    )
+
+    initial_molecule: Molecule = Field(None, description="Input Molecule object")
+    optimized_molecule: Molecule = Field(None, description="Optimized Molecule object")
+
+    # TODO: Discuss with Evan if these go here
```
espottesmith, line 54: I'm not sure that they need to go in the OutputDoc specifically, but I do want them to be part of the drone/parsing/task doc creation! The build pipeline - especially the molecule association builder - is much more efficient if it can query by hashes rather than by formula alone. So it'd be good if tasks could have these from the beginning.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+    #     None,
+    #     description="Weisfeiler Lehman (WL) graph hash using the atom coordinates as the graph node attribute.",
+    # )
+
+    # last_updated: datetime = Field(
+    #     None,
+    #     description = "Timestamp for the most recent calculation for this QChem task document",
+    # )
+
+    final_energy: float = Field(
```
espottesmith, line 69: This is probably a fine start, but it would be nice to have more output data on the top level. Currently, most of the data is stuck in calcs_reversed; that's fine (at least it's somewhere and accessible to the user), but makes accessing it a bit annoying.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+)
+
+from emmet.core.qchem.task import QChemStatus
+
+__author__ = "Rishabh D. Guha <rdguha@lbl.gov>"
+logger = logging.getLogger(__name__)
+
+# class QChemObject(ValueEnum):
+# Not sure but can we have something like GRAD and HESS
+# as QChem data objects
```
espottesmith, line 33: Maybe something we can discuss? Is the idea of this enum that you would have all of the different possible output files as possible enum values?

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+
+    initial_molecule: Dict[str, Any] = Field(
+        None, description="input molecule geometry before the QChem calculation"
+    )
+
+    # parameters: Dict[str, Any] = Field(
+    #     None, description = "Parameters from a previous QChem calculation."
+    # )
+
+    charge: int = Field(None, description="The charge of the input molecule")
```
espottesmith, line 49: If charge is top-level, spin multiplicity should be as well

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+
+__author__ = "Rishabh D. Guha <rdguha@lbl.gov>"
+logger = logging.getLogger(__name__)
+
+# class QChemObject(ValueEnum):
+# Not sure but can we have something like GRAD and HESS
+# as QChem data objects
+
+
+class CalculationInput(BaseModel):
```
espottesmith, line 36: If this is supposed to mimic/map onto the pymatgen QCInput, I think you're missing some sections (e.g. nbo, pcm_nonels, etc.)

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+
+    cds_gradients: Union(Dict[str, Any], np.ndarray) = Field(
+        None, description="The parsed CDS gradients."
+    )
+
+    dipoles: Dict[str, Any] = Field(
+        None, description="The associated dipoles for Mulliken/RESP/ESP charges"
+    )
+
+    gap_info: Dict[str, Any] = Field(
```
espottesmith, line 207: I remember saying at one point that I had no interest in including HOMO-LUMO in MPcules,  but that if someone else wanted to, I wouldn't stop them. Now you've started the work, and I feel myself tempted to go back on my work and fight you.

But I won't. I will stay strong, even though the HOMO-LUMO beast is evil.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+            pcm=qcinput.pcm,
+            solvent=qcinput.solvent,
+            smx=qcinput.smx,
+            vdw_mode=qcinput.vdw_mode,
+            scan_variables=qcinput.scan,
+            van_der_waals=qcinput.van_der_waals,
+        )
+
+
+class CalculationOutput(BaseModel):
```
espottesmith, line 130: Again, we should probably have a conversation about what details should be on the top level. Also, is this meant to represent the entire QCOutput, or just the main "output" data that's in a task document?

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+            nbo_data=qcoutput.data["nbo_data"],
+            frequencies=qcoutput.data["frequencies"],
+            frequency_modes=qcoutput.data.get(["frequency_mode_vectors"], []),
+            final_energy=qcoutput.data["final_energy"],
+            enthalpy=qcoutput.data["enthalpy"],
+            entropy=qcoutput.data["entropy"],
+            scan_energies=qcoutput.data.get(["scan_energies"]),
+            scan_geometries=qcoutput.data["optimized_geometries"]
+            if qcoutput.data["scan_energies"]
+            else None,
```
espottesmith, line 241: I don't know what the current preferred style is, but I'd rather spend a few extra lines and have this logic separate.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+            pcm_gradients=qcoutput.data["pcm_gradients"][0],
+            cds_gradients=qcoutput.data["CDS_gradients"][0],
+            dipoles=qcoutput.data["dipoles"],
+            gap_info=qcoutput.data["gap_info"],
+        )
+
+    # TODO What can be done for the trajectories, also how will walltime and cputime be reconciled
+
+
+class Calculation(BaseModel):
```
espottesmith, line 254: No calcs_reversed? Is there any place where ALL of the data is going to be stored?

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+    """Full QChem calculation inputs and outputs."""
+
+    dir_name: str = Field(None, description="The directory for this QChem calculation")
+    qchem_version: str = Field(
```
espottesmith, line 258: Important. This should have been included in task documents from the jump.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+        None, description="QChem version used to perform the current calculation"
+    )
+    has_qchem_completed: Union[QChemStatus, bool] = Field(
```
espottesmith, line 261: Completed and completed successfully are two different things. Which do you mean to store here?

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+    )
+    task_name: str = Field(
+        None,
+        description="Name of task given by custodian (e.g. opt1, opt2, freq1, freq2)",
+    )
+    output_file_paths: Dict[str, str] = Field(
+        None,
+        description="Paths (relative to dir_name) of the QChem output files associated with this calculation",
+    )
+    level_of_theory: LevelOfTheory = Field(
```
espottesmith, line 281: On the calculation level, you could break this up - functional, basis set, solvent method, solvent, etc. Or you could have functional/basis/solventmethod/etc. AND the full level of theory / lot_solvent. But just having lot feels wrong at this point.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+    )
+
+    @classmethod
+    def from_qchem_files(
+        cls,
+        dir_name: Union[Path, str],
+        task_name: str,
+        qcinput_file: Union[Path, str],
+        qcoutput_file: Union[Path, str],
+        store_energy_trajectory: bool = False,
```
espottesmith, line 301: You should probably be able to tell if you need a trajectory based on whether or not it's an optimization job. Seems to me that opt/tsopt jobs should just about always have optimization trajectories associated.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+        output_doc = CalculationOutput.from_qcoutput(qcoutput)
+
+        has_qchem_completed = (
+            QChemStatus.SUCCESS
+            if qcoutput.data.get("completion", [])
+            else QChemStatus.FAILED
+        )
+
+        if store_energy_trajectory:
+            print("Still have to figure the energy trajectory")
```
espottesmith, line 355: Lol

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+    Dict[str, Any]
+        The filenames of the calculation outputs for each QChem task, given as a ordered dictionary of::
+
+            {
+                task_name:{
+                    "qchem_out_file": qcrun_filename,
+                },
+                ...
+            }
+    If there is only 1 qout file task_name will be "standard" otherwise it will be the extension name like "opt_0"
```
espottesmith, line 398: Seems reasonable.

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,437 @@
...
+    """
+    path = Path(path)
+    task_files = OrderedDict()
+    print(Path)
+
+    in_file_pattern = re.compile(r"^(?P<in_task_name>mol\.qin(?:\..+)?)\.gz$")
```
espottesmith, line 404: Probably worth accepting *.inp and *.out, since those are actually Q-Chem's preferred input and output suffixes (Sam did his own thing).

emmet-core/emmet/core/qchem/task.py
```diff
@@ -29,7 +29,7 @@ class QChemStatus(ValueEnum):
     Q-Chem Calculation State
     """
 
-    SUCESS = "successful"
+    SUCCESS = "successful"
```
espottesmith, line 32: Word.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+    lev_theory: LevelOfTheory = Field(
+        None, description="Level of theory used in the qchem calculation"
+    )
+
+    task_type: TaskType = Field(
+        None,
+        description="The type of the QChem calculation : optimization, single point ... etc.",
+    )
+
+    tags: Union[List[str], None] = Field(
```
espottesmith, line 148: Tags needs to be Any. Some people have lists of tags. Sam and I use dicts.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+        title="Input Structure",
+        description="Input molecule and calc details for the QChem calculation",
+    )
+
+    prev_rem_params: Dict = Field(
+        None,
+        description="Parameters from a previous qchem calculation in the series",
+    )
+
+    lev_theory: LevelOfTheory = Field(
```
espottesmith, line 139: Do not like this name. Should be level_of_theory to be consistent with elsewhere in emmet.

espottesmith: Also, as mentioned elsewhere, this can be broken down more granularly (e.g. to functional, basis, solvent method)

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+        None,
+        description="Dictionary with the implicit solvent model and respective parameters",
+    )
+
+    lot_solvent: str = Field(
+        None,
+        description="A string representation with the combined information of the solvemt used and the level of theory",
+    )
+
+    custom_smd: str = Field(
```
espottesmith, line 162: solvent incorporates all custom_smd information.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+        None, description="Parameter string for SMD implicit solvent model"
+    )
+
+    special_run_type: str = Field(
+        None, description="Special workflow name (if applicable)"
+    )
+
+    smiles: str = Field(
```
espottesmith, line 170: If we're going to store SMILES, we should also store InChI.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+    )
+
+    job: dict = Field(
+        None,
+        title="Custodian Job Data",
+        description="Job data logged by custodian.",
+    )
+
+
+# AnalysisDoc? Is there a scope for AnalysisDoc in QChem?
```
espottesmith, line 226: I'm not sure. Is there a VASP analysis doc? If so, what's its purpose?

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+
+    dir_name: str = Field(None, description="The directory for this QChem task")
+
+    state: QChemStatus = Field(None, description="State of this QChem calculation")
+
+    calcs_reversed: List[Calculation] = Field(
+        None,
+        title="Calcs reversed data",
+        description="Detailed data for each QChem calculation contributing to the task document.",
+    )
```
espottesmith, line 242: Okay, here we go. Nice. But if Calculation is meant to represent the full data in calcs_reversed, then that means that right now, you're throwing out a TON of information. Calculation will need to be seriously expanded. There's a reason that I previously just used dicts for calcs_reversed (other than me being lazy): there's just a lot of fields.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+        None, description="Final optimized/output molecule from this task"
+    )
+
+    task_type: Union[CalcType, TaskType] = Field(
+        None, description="the type of QChem calculation"
+    )
+
+    orig: Dict[str, Any] = Field(
+        {}, description="Summary of the original Q-Chem inputs"
+    )
```
espottesmith, line 254: Orig is basically an inputdoc as well. It's just the original input. So if input is an InputDoc, I'd think orig should be as well?

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+        None,
+        description="The input molecule and calc parameters used to generate the current task document.",
+    )
+
+    output: OutputDoc = Field(
+        None,
+        description="The exact set of output parameters used to generate the current task document.",
+    )
+
+    # TODO: Implement entry dict
```
espottesmith, line 266: Please do.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+
+    custodian: List[CustodianDoc] = Field(
+        None,
+        title="Calcs reversed data",
+        description="Detailed custodian data for each QChem calculation contributing to the task document.",
+    )
+
+    critic2: Dict[str, Any] = Field(
+        None, description="Outputs from the critic2 calculation if performed"
+    )
```
espottesmith, line 276: Going forward, there's really no reason that this needs to be top-level. It could just be within output. But it seems like you're trying to stick with the current (e.g. atomate1) format, in which case of course this should stay.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+
+    custom_smd: Union(str, Dict[str, Any]) = Field(
+        None,
+        description="The seven solvent parameters necessary to define a custom_smd model",
+    )
```
espottesmith, line 281: Again, custom_smd is really unnecessary if you're storing solvent elsewhere, but if you're just trying to stick with the current format, that's totally fine.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+            "molecule": calcs_reversed[0].output.molecule,
+            "composition": calcs_reversed[0].output.molecule.composition,
+            "formula": calcs_reversed[
+                0
+            ].output.formula.composition.aplhabetical_formula,
+            "energy": calcs_reversed[0].output.final_energy,
+            "output": calcs_reversed[0].output.as_dict(),
+            "critic2": calcs_reversed[
+                0
+            ].output.critic,  # TODO: Unclear about orig_inputs
```
espottesmith, line 426: Is critic2 guaranteed to be in the first calculation? Seems to me that you should iterate over calculations.

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+            "task_id": task_id,
+            "charge": calcs_reversed[0].output.molecule.charge,
+            "spin_multiplicity": calcs_reversed[0].output.molecule.spin_multiplicity,
+            "level_of_theory": calcs_reversed[-1].input.lev_theory,
+            "solvent": calcs_reversed[-1].input.solv_spec,
+            "lot_solvent": calcs_reversed[-1].input.lot_solv_combo,
+            "custom_smd": calcs_reversed[-1].input.custom_smd,
+            "task_type": calcs_reversed[-1].input.task_spec,
+            "calc_type": calcs_reversed[-1].input.calc_spec,
+            "tags": calcs_reversed[-1].input.tags,
```
espottesmith, line 416: Why are these grabbed from the last (first) calc?

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,609 @@
...
+            "critic2": calcs_reversed[
+                0
+            ].output.critic,  # TODO: Unclear about orig_inputs
+            "last_updated": calcs_reversed[0].output.last_updated,
+        }
+
+        return entry_dict
+
+
+def get_uri(dir_name: Union[str, Path]) -> str:
```
espottesmith, line 433: Cool.

espottesmith (commented):
Seems like a really good start. A few small style or name requests. But mostly, I just have questions/concerns about what things are stored where. I want to make sure that data isn't lost, that we're grabbing everything that we care about, and that that data is organized in a logical way.

emmet-core/emmet/core/qc_tasks.py
Andrew-S-Rosen: Technically, I suppose it should be type `MSONable`?

emmet-core/emmet/core/qc_tasks.py
Andrew-S-Rosen: Indeed [there is](https://github.com/materialsproject/emmet/blob/8df1e350d30aed8bdbbc9205375a79c331d97c32/emmet-core/emmet/core/tasks.py#L220). It is to report on things such as the change in volume over the course of the relaxation and the max force. The idea is that they are parameters that users want to query by to determine if the calculations seem sketchy, e.g. the volume changed too much or the force is too high.

emmet-core/emmet/core/qchem/calculation.py
rkingsbury: Agree in principle, but if we're just duplicating atomate1's [`generate_doc`](https://github.com/hackingmaterials/atomate/blob/6327b07ea1ae85fbbe29772777a80dfe4a94dc7b/atomate/qchem/drones.py#L146) method, only `molecule` is stored at the top level. That method overrides the molecule charge, too.

Would it be enough to store an entire `Molecule` object under `initial_molecule`, that will carry charge and spin information with it?

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,669 @@
...
+    dir_name: str = Field(None, description="The directory for this QChem calculation")
+    qchem_version: str = Field(
+        None, description="QChem version used to perform the current calculation"
+    )
+    has_qchem_completed: Union[QChemStatus, bool] = Field(
+        None, description="Whether QChem calculated the calculation successfully"
+    )
+    input: CalculationInput = Field(
+        None, description="QChem input settings for the calculation"
+    )
```
rkingsbury, line 308: Is this meant to be the input document? If so, please clarify the docstsring for consistency with `output`, which says explicitly "calculation output document`

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,669 @@
...
+    #
+    #     string = f"DIELECTRIC={round(dielectric, 2)},RHO={round(rho, 4)}"
+    #     for name, (piece, digits) in {"A": (a, 6), "B": (b, 6), "C": (c, 1), "D": (d, 3),
+    #                                   "SOLVRHO": (solvrho, 2), "GAMMA": (gamma, 1)}.items():
+    #         if piece is None:
+    #             piecestring = "NONE"
+    #         else:
+    #             piecestring = f"{name}={round(float(piece), digits)}"
+    #         string += "," + piecestring
+    #     return string
```
rkingsbury, line 597: I believe this is all merged now!

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,616 @@
...
+        None,
+        title="Custodian Corrections",
+        description="List of custodian correction data for calculation.",
+    )
+
+    job: Optional[Dict[str, Any]] = Field(
+        None,
+        title="Custodian Job Data",
+        description="Job data logged by custodian.",
+    )
```
rkingsbury, line 218: Is there an equivalent in VASP? Hopefully a lot of boilerplate you could borrow if so

emmet-core/emmet/core/qchem/calculation.py
rkingsbury: I've also seen `qcin` and `qcout`. Maybe the input and output extensions to search for should be kwargs? (list type, defaulting to just `[".qin"]`?

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,669 @@
...
+    Dict[str, Any]
+        The filenames of the calculation outputs for each QChem task, given as a ordered dictionary of::
+
+            {
+                task_name:{
+                    "qchem_out_file": qcrun_filename,
+                },
+                ...
+            }
+    If there is only 1 qout file task_name will be "standard" otherwise it will be the extension name like "opt_0"
```
rkingsbury, line 451: According to the docstring, this method is only supposed to return output files, but the regex pattern `in_file_pattern` will match _input_ files

emmet-core/emmet/core/qc_tasks.py
```diff
@@ -0,0 +1,616 @@
...
+
+        critic2 = {}
+        custom_smd = {}
+        calcs_reversed = []
+        for task_name, files in task_files.items():
+            if task_name == "orig":
+                continue
+            else:
+                calc_doc = Calculation.from_qchem_files(
+                    dir_name, task_name, **files, **qchem_calculation_kwargs
```
rkingsbury, line 329: The `**files` construction might be brittle, because you are depending on 1) there only being two keys in each task and 2) them being ordered correctly. I suggest explicitly setting the input and output file kwargs according to the respective keys

emmet-core/emmet/core/qchem/calculation.py
```diff
@@ -0,0 +1,669 @@
...
+            has_qchem_completed=has_qchem_completed,
+            completed_at=completed_at,
+            input=input_doc,
+            output=output_doc,
+            output_file_paths={
+                k.lower(): Path(v)
+                if isinstance(v, str)
+                else {k2: Path(v2) for k2, v2 in v.items()}
+                for k, v in output_file_paths.items()
+            },
```
rkingsbury, line 417: Since each `Calculation` document describes one specific Q-Chem calc, with a particular input and output file, it seems odd to me to store the complete list of output file paths here. Doesn't that belong at the Task Document level?

rkingsbury (commented):
Wow, thanks for the painstaking work on this @rdguha1995 ! I went through and made some review comments which hopefully make sense. Feel free to contradict me though; I haven't had my head in Q-Chem infrastructure in a little while so some may be ill-informed.

rkingsbury:
> This is very much in line with Alex's vision of constructing TaskDoc's for VASP using separate models for Input and Output

Can you help me understand the overlap / relation to the existing `emmet.core.qchem.TaskDocument`? If I'm understanding correctly, the intention here is to
1. Update that class to have separate Input and Output documents, and
2. add a `from_directory` method which the existing `emmet` class lacks

Your `TaskDoc` and the existing `TaskDocument` are not meant to co-exist long term, right?

rdguha1995:
Great catch @rkingsbury . The `emmet.core.qchem.TaskDocument` was the one Evan crafted to generate a `TaskDocument` from calculations already parsed through the `atomate` drone. In the short term I didn't want to replace the existing code in the fear of breaking something. So I developed this parallel `TaskDoc` class which generates the document from the `Calculation` `pydantic` model we have devised in this PR. In the long term, only 1 of these will remain.

PR #824, opened 2023-09-12
Add `is_hubbard` and `hubbards` to entry.parameters

mkhorton:
A small patch for backwards-compatibility with the new `TaskDoc`; this is required by the compatibility scheme. Not sure how we missed this before, I thought I'd added! Thanks to Robert Pinsler for notification.

Should be fine to merge/release as-is, however longer term a document model for the Entry itself might be useful that specifies these fields.


mkhorton:
Btw, I noticed the last PR I had which was merged had a release label applied but a release was not made.

munrojm:
Thanks for the fix. I'll tag this as well, and it should release properly this time if the linting stage passes. You will have to run ruff locally to bring it in-line.

mkhorton:
Ah, gotcha, thanks. Hopefully should be good now. Was a one length / black issue with previous commit.

munrojm (approved):

PR #837, opened 2023-09-27
Remove the pymatgen<=2023.9.10 version pin from `emmet-core`

Andrew-S-Rosen:
See https://github.com/materialsproject/pymatgen/releases/tag/v2023.9.10, in which we removed the pydantic pin. So, given that emmet-core now supports pydantic 2, I have updated it to `pymatgen>=2023.9.10`.

### DIFF
--- a/emmet-core/setup.py
+++ b/emmet-core/setup.py
@@ -15,7 +15,7 @@ setup(
     },
     include_package_data=True,
     install_requires=[
-        "pymatgen>=2023.9.10",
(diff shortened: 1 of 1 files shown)


munrojm:
I added the pin today due to an issue with using `OBAlign` from openbabel. @janosh fixed this, but we will have to wait for next release.

Andrew-S-Rosen:
Makes sense! Thought it was included because of pydantic 👍

PR #853, opened 2023-10-12
Fix `TaskDoc.set_entry`

janosh:
Fixes `atomate2` defect workflow test. See https://github.com/materialsproject/atomate2/pull/548#issuecomment-1758572768 for details.

### DIFF
--- a/emmet-api/app.py
+++ b/emmet-api/app.py
@@ -1,6 +1,8 @@
 import time
+
 start = time.perf_counter()
 import logging
+
 logging.getLogger("uvicorn.access").handlers = []
 from asgi_logger import AccessLoggerMiddleware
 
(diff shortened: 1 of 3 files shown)


janosh:
Not sure why `self.get_entry` was required in the 1st place since `task_id` is optional in `TaskDoc.get_entry`:

https://github.com/materialsproject/emmet/blob/6ac53ecb4b34a1e508a32cb807fa4c864c991086/emmet-core/emmet/core/tasks.py#L595

jmmshn:
Also, can we add a test case to check that the `entry` field is populated?

janosh:
> Also, can we add a test case to check that the entry field is populated?

Done in daa0034.

PR #857, opened 2023-10-12
`MoleculeMetadata` set model config `allow='extra'`

janosh:
Fixes breaking `test_cclib_taskdoc` in https://github.com/materialsproject/atomate2/pull/548.

[Failing lines that require `allow='extra'`](https://github.com/materialsproject/atomate2/blob/302bed8b1e4fd68aec3ffbfa92302780c58b9902/tests/common/schemas/test_cclib.py#L91-L92).

2nd fix needed for `atomate2` tests following #853.

### DIFF
--- a/emmet-api/app.py
+++ b/emmet-api/app.py
@@ -30,9 +30,6 @@ app.add_middleware(
     AccessLoggerMiddleware,
     format='%(h)s %(t)s %(m)s %(U)s?%(q)s %(H)s %(s)s %(b)s "%(f)s" "%(a)s" %(D)s %(p)s %({x-consumer-id}i)s',
 )
-app.add_middleware(
-    CORSMiddleware,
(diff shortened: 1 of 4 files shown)


munrojm:
The issue with this is that the metadata model is inherited by a lot of other documents. This effectively destroys validation for them. Can we not just enable extra fields on `TaskDocument` like we do with the VASP equivalent (https://github.com/materialsproject/emmet/blob/485638ec7bb688462eed9c35d5c443e22a84941f/emmet-core/emmet/core/tasks.py#L310)?

janosh:
@munrojm That works too. I now test that extra fields are ignored and will set `extra='allow'` [here](https://github.com/materialsproject/atomate2/blob/302bed8b1e4fd68aec3ffbfa92302780c58b9902/src/atomate2/common/schemas/cclib.py#L29) instead.

PR #869, opened 2023-10-20
Add `run_ddec6` option to `Calculation.from_vasp_files`

janosh:
This PR adds the functionality needed to automatically run a DDEC6 charge analysis after any VASP calculation with `atomate2`. See https://github.com/materialsproject/atomate2/pull/587.

@chiang-yuan Should the doc string contain some instructions on how to set `DDEC6_ATOMIC_DENSITIES_DIR` and where to get those densities?

### DIFF
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -661,7 +661,8 @@ class Calculation(BaseModel):
         run_ddec6 : bool | str = False
             Whether to run DDEC6 on the charge density. If a string, it's interpreted
             as the path to the atomic densities directory. Can also be set via the
-            DDEC6_ATOMIC_DENSITIES_DIR environment variable.
+            DDEC6_ATOMIC_DENSITIES_DIR environment variable. The files are available at
+            https://sourceforge.net/projects/ddec/files.
         strip_dos_projections
             Whether to strip the element and site projections from the density of
             states. This can help reduce the size of DOS objects in systems with many
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -3,13 +3,14 @@
 # mypy: ignore-errors
 
 import logging
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
-from datetime import datetime
 
 import numpy as np
 from pydantic import BaseModel, Extra, Field
 from pymatgen.command_line.bader_caller import bader_analysis_from_path
+from pymatgen.command_line.chargemol_caller import ChargemolAnalysis
 from pymatgen.core.lattice import Lattice
 from pymatgen.core.structure import Structure
 from pymatgen.core.trajectory import Trajectory
@@ -18,6 +19,7 @@ from pymatgen.electronic_structure.core import OrbitalType
 from pymatgen.electronic_structure.dos import CompleteDos, Dos
 from pymatgen.io.vasp import (
     BSVasprun,
+    Kpoints,
     Locpot,
     Outcar,
     Poscar,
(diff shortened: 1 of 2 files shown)


chiang-yuan:
Yes I totally agree we should have that. It didn't require me long time to figure out what are the files and where to get them. They are in the author's repo [here](https://github.com/berquist/chargemol/tree/master/atomic_densities). Perhaps we should keep a copy/snapshot of all the files in case the authors are not actively maintaining the repo or just deleted the files in the future?

janosh:
> Perhaps we should keep a copy/snapshot of all the files in case the authors are not actively maintaining the repo or just deleted the files in the future?

How big are the files gzipped? If they're < 50KB, we could keep a compressed copy in `pymatgen` and try downloading those files on the fly when users run `chargemol`.

Andrew-S-Rosen:
@janosh and @chiang-yuan: 

For your own benefit, please read the email I sent to Tom Manz below a long time ago, as it will be relevant for you if you use this feature.

> Dear Dr. Manz,

> I have run DDEC6 via Chargemol without issue for several years now. However, I wanted to bring to your attention some strange behavior when running the parallel version of Chargemol on certain machine architectures. I understand that this behavior may be difficult to reproduce, but I share it here with you for completeness anyway.

> Summary:
> When using the parallel version of Chargemol (i.e. Chargemol_09_26_2017_linux_parallel) on the head node of NERSC's Cori machine (Intel Xeon Processor E5-2698), the predicted partial atomic charges (and other properties) are erratic on simple test systems. For instance, for the NaCl test case provided with Chargemol, the partial charge on Na is predicted to be -1.9, whereas with the serial executable it is +0.843 (as expected). Rerunning the parallel code yields very different partial charges on most runs, whereas the serial version consistently yields +0.843.

> Importantly, I do not have this issue on all machine architectures, which has made it difficult to pinpoint the issue. Running the exact same setup on Intel Xeon nodes on a separate machine (one I've historically used for such calculations) yields the expected partial charges for NaCl every time regardless of the executable.

> The Data:
> I have attached to this email the exact files I used to run Chargemol as well as the outputs. The "serial" and "parallel" folders refer to the serial and parallel Linux binaries of Chargemol, respectively.

> Perhaps you may have some insight as to the root cause. If so, please let me know if there is any troubleshooting I can do to help out further.
[nacl_tests.zip](https://github.com/materialsproject/emmet/files/13059144/nacl_tests.zip)

The response from Tom:

> It seems like it is due to some issues of the system architecture and/or the compiler that was used on that system. You may need to recompile the code on that system using a few different versions of the GNU fortran compiler to see if one of them solves the issue. I've noticed that sometimes a particular version of the GCC fortran compiler works when another will produce errors when the code runs. So, my recommendation is to explore some options with different GCC fortran compiler versions.

> If you transferred the code from one system to another without recompiling it, this sometimes works if the two systems have close enough architectures. But, it may be in your case that the architecture is different enough that if the code was compiled on a vastly different architecture that it needs to be recompiled.

> So, in summary ... I think your issue sounds like a compiler issue.

Andrew-S-Rosen:
@chiang-yuan: That is not the repo for DDEC6! It is here: https://sourceforge.net/projects/ddec/files/

chiang-yuan:
@Andrew-S-Rosen Thanks for the catch. You are definitely spot on, although the time stamps on two side are the same... I should be more careful about this...

chiang-yuan:
@janosh I compressed the whole directory into tar.gz and it is about 2.3M. Sounds like it is not very small?

chiang-yuan:
@Andrew-S-Rosen to the information you provided above, I have been compiling ddec charges for all MP computed entries. As far as I've seen for Na, there is no such erroneous entry with Na of -1.9. Very few of them are mildly negatively charged (roughly around or above -0.5) but those might be charge sloshing in pure metal.

Andrew-S-Rosen:
Good good! That's consistent with my experience too but just wanted you to be alerted. I only saw it on the head node on Cori, not the compute node, so who knows. And of course, Cori is gone now.

janosh:
2.3M is not great but still ok. If they've been hosted reliably for a long time where they are now, we can also pull from there.

PR #874, opened 2023-10-28
Add   `summary_stats` attr to `emmet.core.vasp.calculation.PotcarSpec`

esoteric-ephemera:
For consistency with pymatgen, which now includes the ability to validate POTCARs without hashing (see [PMG PR #3351](https://github.com/materialsproject/pymatgen/pull/3351)), added a `summary_stats` attribute to `emmet.core.vasp.calculation.PotcarSpec`.

`summary_stats` is just a dict with structure
```py
self._summary_stats = {
    "keywords": {
        "header": list[str],
        "data": list[str]
    },
    "stats": {
        "header": dict[float],
        "data": dict[float]
    }
}
```

This will help in adding new the POTCAR validation scheme to emmet.

### DIFF
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -65,8 +65,8 @@ class PotcarSpec(BaseModel):
 
     titel: Optional[str] = Field(None, description="TITEL field from POTCAR header")
     hash: Optional[str] = Field(None, description="md5 hash of POTCAR file")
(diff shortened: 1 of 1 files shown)


janosh:
Would be good to add a test for the `summary_stats` attribute in [emmet-core/tests/test_calculation.py](https://github.com/materialsproject/emmet/blob/1e65eebdda425ab406b021bd609b2c90ea26487a/emmet-core/tests/test_calculation.py).

PR #883, opened 2023-11-07
Fix custodian json import issue

JaGeo:
Closes https://github.com/materialsproject/atomate2/issues/604. I would be happy if this could get merged soon.


JaGeo:
@munrojm, would it be possible for you to have a look at this?

utf:
Hi @JaGeo, thanks for this report.

A couple of thoughts:
- Is this issue happening because the json parsing is using monty `loadfn` which is deserialising the objects? One fix would be to just use `json.load`. This will probably be quicker and less error prone.
- To avoid more issues down the line, we could also change type hinting to `Optional[Any]` as per this comment: https://github.com/materialsproject/atomate2/issues/604#issuecomment-1800037935

JaGeo:
> * Is this issue happening because the json parsing is using monty `loadfn` which is deserialising the objects? One fix would be to just use `json.load`. This will probably be quicker and less error prone.

I assume so. I think, however, that it does not happen during the initial reading in but rather at a later point as the initial reading suppresses the usage of cls. 
https://github.com/materialsproject/emmet/blob/be40a1f58c8911148df2d33874f0412855b0a7ee/emmet-core/emmet/core/tasks.py#L785

munrojm:
@JaGeo sorry for the delay. This looks fine to me. I would prefer to be a bit more explicit with `Optional[Union[dict,VaspJob]]` as you have it.

JaGeo:
@munrojm Thank you so much!

PR #884, opened 2023-11-10
fix custodian import

JaGeo:
@munrojm I somehow overlooked that custodian is not installed by emmet... Therefore, importing custodian fails. I would assume we do not want custodian as a dependency just for type checking.

Really sorry! I am currently using the mp-api and it is failing because of my latest change.


rkingsbury:
I just noticed this too!

mattmcdermott:
Referenced downstream here: https://github.com/materialsproject/api/issues/861

Issue #872, opened 2023-10-26
[Feature Request]: Better handling of parsed trajectory in VASP calculations


gpetretto:
### Problem

The current calculation model used for MD simulations in atomate2 relies on the `store_trajectory` option in the `Calculation` model.
I think that there are some problems associated with the current handling of the trajectory. Those related to emmet concern the amount of data that is stored. In fact, for each frame of the `Trajectory` object the whole `IonicStep` is stored as well
https://github.com/materialsproject/emmet/blob/ebbe8d5124c9e2bd5b743698836c7a17c6d693a5/emmet-core/emmet/core/vasp/calculation.py#L732-L734

Considering that this feature is used to store the output of (potentially long) MD simulation, this has some issues:
1) each ionic step contains the `Structure` of the step, so each Structure is essentially stored twice in the Trajectory. It seems that the structure stored in the IonicStep does not include any additional property, so it is entirely redundant.
2) for each step all the SCF steps are also stored as well, which amount to a considerable amount of data being stored in the DB
3) Some additional information that is only available for MD simulation is not considered. In particular I am referring to the temperature, that would be interesting to have for post-processing analysis.

The first two points are not blocking issues, but when storing the output of long simulations they will probably have a big impact on the size required for the stored data. Combined, they will probably require roughly 3x the space required if only the structure and the final properties would be stored for each step.

The parsing of large MD trajectory may also be demanding in terms of resources and in general I believe that this could be optionally handled by extracting the data from the new vasp hd5 output file, rather than from the vasprun.xml. So this issue might be linked to the one I opened in atomate2: https://github.com/materialsproject/atomate2/issues/515

Also tagging @utf and @mjwen as this may impact the MD flow in atomate2.

### Proposed Solution

I am aware of the fact that if the trajectory is stored its content covers also the one of the `ionic_steps` in the `CalculationOutput` model. But I think that it would be reasonable to have the following changes

* Remove the Structure from the frame properties to avoid redundancy. This should be easy to do and I don't see any downside. The `structure` attribute is even optional in `IonicStep`
* I don't see a strict need for the electronic steps to be present in a long MD simulation, but there are probably cases where this might be needed. I would thus propose to have the store_trajectory attribute as a multi valued flag, that would allow to store the full `IonicStep` (except for the Structure) or a subset of the data. 
  To preserve backward compatibility this could be defined as
  ```python
  store_trajectory: bool | Literal["partial"] = False,
  ```
[...]

PR #886, opened 2023-11-13
Improve trajectory handling for MD

gpetretto:
Following what discussed in #872, I have tried to implement an additional option for `store_trajectory` and to add the temperature parsed from the OSZICAR. 

For the temperature, the code will attempt to parse it if `store_trajectory` is not False, but there are different ways of doing this. For example the temperature could have been stored in a different location in the model, or its parsing could have been activated by a new option. I preferred to keep it as minimal as possible and see if there are specific requests for emmet or atomate2 compatibility (@utf and @mjwen).

### DIFF
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -21,6 +21,7 @@ from pymatgen.io.vasp import (
     BSVasprun,
     Kpoints,
     Locpot,
+    Oszicar,
     Outcar,
     Poscar,
     Potcar,
@@ -612,6 +613,7 @@ class Calculation(BaseModel):
         contcar_file: Union[Path, str],
         volumetric_files: List[str] = None,
         elph_poscars: List[Path] = None,
+        oszicar_file: Optional[Union[Path, str]] = None,
         parse_dos: Union[str, bool] = False,
         parse_bandstructure: Union[str, bool] = False,
         average_locpot: bool = True,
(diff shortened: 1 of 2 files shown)


mjwen:
Thanks @gpetretto ! 

I've no specific request. This looks great.

emmet-core/emmet/core/vasp/calculation.py
```diff
@@ -618,7 +622,7 @@ def from_vasp_files(
         strip_bandstructure_projections: bool = False,
         strip_dos_projections: bool = False,
         store_volumetric_data: Optional[Tuple[str]] = None,
-        store_trajectory: bool = False,
+        store_trajectory: Union[bool, str] = False,
```
munrojm, line 625: Instead of mixing `bool` and `str` inputs to support a third "partial" option, I would vote change this to be an enum value or string representation of an enum value. That way the type and options are more well-defined.

munrojm:
This looks great! Posted one comment, but I think this is a nice addition.

emmet-core/emmet/core/vasp/calculation.py
gpetretto: I switched to the Enum. Passing the string value is also fine. Let me know if you have other suggestions or better naming conventions.

munrojm: Looks fine to me, let me know when you are happy to have it merged.

gpetretto: If it is fine, you can proceed to merge. Thanks!

PR #891, opened 2023-11-15
Cli backup expansion

tsmathis:
**This PR expands on the `backup` function of the emmet cli by:**

1. Modifying the logic for verifying  the validity of block launchers that have already been archived
    - the default behavior of the `--check` flag remains unchanged, where the entire `.tar` archive for a block launcher will be checked as a single unit and hard fail if the verification fails 
    - there is now an additional `--exhaustive` flag that  will check the validity of each individual launcher in the `.tar` archive for a block and then only add launchers that pass verification into a "safe to remove" list which will then be cleaned up (if `clean` is active). This "soft fail" method 1) prevents `--clean`ing of any launchers that fail verification, while 2) allowing verification/cleaning to continue for launchers that do pass. Common failures have been due to local file inconsistencies on `cfs`, i.e., launcher was cleaned previously, user moved/modified files after initial archiving. 

Example output from the new `--exhaustive` check when paired with `--clean`:
Screenshot 2023-11-15 at 12 01 00 PM

2. Adding an additional `--tar` flag that compresses whatever is left in a block launcher on disk after archiving, checking, and cleaning. `--tar` is only ever invoked if `--clean` is also active. 

**This PR also adds a new `tasks` functionality that supplements the `backup` functionality of the cli: `survey`**

- `emmet tasks survey` will search through a given directory (and all sub-dirs) to find any directories containing VASP files and put together a report-style log file
- `survey`'s functionality comes from the `launcher_finder` bash script, which will be installed on the user's path when `emmet-cli` is `pip` installed, and can be invoked directly for convenience
- Similar in spirit to `backups` (recursive search), but does not submit `backup` jobs. `survey` will also find VASP files that need to be `--reorg`ed or that have been quarantined due to failing validation

`survey` is mainly intended for use as a discovery tool  to avoid manual searching. For example, when given a user's directory for archival where we have no idea about the contents (for MP, e.g., NERSC user that left years ago and we have nothing to go on).

Small example output from `survey`:

Screenshot 2023-11-15 at 11 44 14 AM


tschaume (approved):
Wonderful!

PR #917, opened 2024-01-01
Replace `emmet.core.utils.ValueEnum` with std lib `enum.StrEnum`

janosh:
In case I missed something, @munrojm was there a reason not to use `enum.StrEnum`?

I was seeing JSON serialization errors with the `atomate2` `PhononMaker`'s `pydantic` model which don't occur with `enum.StrEnum`.

```py
 json.dump(atomate2_phonon_doc, json_file, cls=MontyEncoder)
```

#### Update

Didn't realize [`StrEnum` is 3.11+ only](https://docs.python.org/3/library/enum.html#enum.StrEnum). We could use a backport package like https://github.com/irgeek/StrEnum.


munrojm:
I have no issues with using a backport package for now.

Andrew-S-Rosen:
@janosh was this at all related to #944?

janosh:
thanks @Andrew-S-Rosen, good to know.

probably still makes sense to define a

```py
class StrEnum(str, Enum)
```

to enforce string values certain enums in `emmet` until oldest python supported is 3.11.

PR #924, opened 2024-01-16
Remove 'model_' from ML doc field names

tsmathis:
The fields `model_name` and `model_version` in the MLDoc clash with the protected namespace `model_` that exists in Pydantic 2. See [here](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.protected_namespaces), where the warning arises to prevent collisions with `BaseModel`s methods.

 Proposed fix is to change `model_name` and `model_version` to just `name` and `model

### DIFF
--- a/emmet-core/emmet/core/ml.py
+++ b/emmet-core/emmet/core/ml.py
@@ -58,10 +58,8 @@ class MLDoc(ElasticityDoc):
     matcalc_version: Optional[str] = Field(
         None, description="Version of matcalc used to generate this document"
     )
-    model_name: Optional[str] = Field(
-        None, description="Name of model used as ML potential."
-    )
-    model_version: Optional[str] = Field(
+    name: Optional[str] = Field(None, description="Name of model used as ML potential.")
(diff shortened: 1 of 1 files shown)


tsmathis:
@munrojm, per our discussion

emmet-builders/emmet/builders/materials/ml.py
```diff
@@ -55,8 +55,8 @@ def __init__(
             model_name = {"chgnetcalculator": "chgnet"}.get(model_name, model_name)
             pkg_name = {"m3gnet": "matgl"}.get(model_name, model_name)
             self.provenance = dict(
-                model_name=model_name,
-                model_version=version(pkg_name),
+                name=model_name,
```
janosh, line 58: is just `model` ok? clearer what it refers to than `name` imo

tsmathis: Makes sense to me! Changing to `model` also doesn't result in any runtime warnings, so should be all good.

PR #925, opened 2024-01-18
added store_onsite_density_matrices option issue atomate2 #671

JonathanSchmidt1:
I added a store_onsite_density_matrices option to `from_vasp_outputs` and `from_vasp_files`. As a result  onsite_density_matrices from the OUTCAR will only be stored in the future if users  specifically ask for  it.
This was added to solve https://github.com/materialsproject/atomate2/issues/671 

I have run the pytest test_calculation.py locally and it passed. I also checked that switching the option on and off in the from_vasp_outputs call in the tests adds/removes the onsite_density_matrices from the outcar.
I am bit surprised that the test passed as I would have expected it to fail when the density matrices are missing but maybe that is because the density matrices of the test cases seem to be empty anyway.

### DIFF
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -455,7 +455,6 @@ class CalculationOutput(BaseModel):
             If not NO, the `ionic_steps` field is left as None.
         store_onsite_density_matrices
             Whether to store the onsite density matrices from the OUTCAR.
-            
         Returns
         -------
             The VASP calculation output document.
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -455,6 +455,7 @@ class CalculationOutput(BaseModel):
             If not NO, the `ionic_steps` field is left as None.
         store_onsite_density_matrices
             Whether to store the onsite density matrices from the OUTCAR.
+            
         Returns
         -------
             The VASP calculation output document.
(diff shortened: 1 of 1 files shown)


utf:
Hi @munrojm, is it possible to take a look at this PR? From our side it is needed to prevent extremely large objects being stored. This field is also automatically excluded by the MP drones.

munrojm:
Looks fine to me. Once the linting is fixed, I will merge and patch release.

emmet-core/emmet/core/vasp/calculation.py
```diff
@@ -547,7 +551,7 @@ def from_vasp_outputs(
                 temp = str(elph_poscar.name).replace("POSCAR.T=", "").replace(".gz", "")
                 elph_structures["temperatures"].append(temp)
                 elph_structures["structures"].append(Structure.from_file(elph_poscar))
-
+
```
utf, line 554:
```suggestion

```

utf (commented):
I think this is the only change needed!

emmet-core/emmet/core/vasp/calculation.py
```diff
@@ -452,7 +453,9 @@ def from_vasp_outputs(
             Different value tune the amount of data from the ionic_steps
             stored in the Trajectory.
             If not NO, the `ionic_steps` field is left as None.
-
+        store_onsite_density_matrices
+            Whether to store the onsite density matrices from the OUTCAR.
+
```
utf, line 458:
```suggestion

```

Sorry, this one too!

PR #937, opened 2024-02-14
Correct typing of TaskDoc.transformations field

esoteric-ephemera:
Objects stored in TaskDoc.transformations can be serialized at init (for example, `pymatgen.alchemy.materials.TransformedStructure`), which leads to pydantic typing errors. This happens in atomate2 when output references are resolved, see discussion in [atomate2 PR #623](https://github.com/materialsproject/atomate2/pull/623).

This PR allows the typing of `TaskDoc.transformations` to be `Any`, and adds tests for the new behavior. 

This would close atomate2 issue [#607](https://github.com/materialsproject/atomate2/issues/607).

### DIFF
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -376,7 +376,7 @@ class TaskDoc(StructureMetadata, extra="allow"):
     icsd_id: Optional[Union[str, int]] = Field(
         None, description="Inorganic Crystal Structure Database id of the structure"
     )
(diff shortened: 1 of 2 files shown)


utf:
Hi @munrojm, once this is merged is it possible to get a new emmet-core released?

Were't hoping this will be the final fix needed to support the latest pydantic version in atomate2.

munrojm:
Yup! Looks good to me.

PR #944, opened 2024-02-21
Fix JSON serialiation of `ValueEnum`s

### LINKED ISSUE #914: Bug: The `.as_dict()` method on `CrystalSystem` is not Monty compatible
### Problem

```python
from emmet.core.symmetry import CrystalSystem
sym = CrystalSystem("Triclinic")
sym.as_dict()
```

This returns the string "Triclinic" rather than a dictionary. This would normally be fine, but the fact that it's a string means that doing `dumpfn` doesn't work on a data structure containing a `CrystalSystem` object.

```python
from monty.serialization import dumpfn

dumpfn(sym, "test.json")
```

Traceback:
```python
------------------------------------------------------------------------
TypeError                              Traceback (most recent call last)
Cell In[47], [line 1](vscode-notebook-cell:?execution_count=47&line=1)
----> [1](vscode-notebook-cell:?execution_count=47&line=1) dumpfn(CrystalSystem("Triclinic"),"test.json")

File [c:\Users\asros\miniconda\envs\zeolites\lib\site-packages\monty\serialization.py:123](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/serialization.py:123), in dumpfn(obj, fn, fmt, *args, **kwargs)
    [121](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/serialization.py:121)     if "cls" not in kwargs:
    [122](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/serialization.py:122)         kwargs["cls"] = MontyEncoder
--> [123](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/serialization.py:123)     fp.write(json.dumps(obj, *args, **kwargs))
    [124](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/serialization.py:124) else:
    [125](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/serialization.py:125)     raise TypeError(f"Invalid format: {fmt}")

File [c:\Users\asros\miniconda\envs\zeolites\lib\json\__init__.py:238](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:238), in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    [232](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:232) if cls is None:
    [233](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:233)     cls = JSONEncoder
    [234](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:234) return cls(
    [235](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:235)     skipkeys=skipkeys, ensure_ascii=ensure_ascii,
    [236](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:236)     check_circular=check_circular, allow_nan=allow_nan, indent=indent,
    [237](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:237)     separators=separators, default=default, sort_keys=sort_keys,
--> [238](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/__init__.py:238)     **kw).encode(obj)

File [c:\Users\asros\miniconda\envs\zeolites\lib\json\encoder.py:199](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:199), in JSONEncoder.encode(self, o)
    [195](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:195)         return encode_basestring(o)
    [196](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:196) # This doesn't pass the iterator directly to ''.join() because the
    [197](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:197) # exceptions aren't as detailed.  The list call should be roughly
    [198](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:198) # equivalent to the PySequence_Fast that ''.join() would do.
--> [199](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:199) chunks = self.iterencode(o, _one_shot=True)
    [200](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:200) if not isinstance(chunks, (list, tuple)):
    [201](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:201)     chunks = list(chunks)

File [c:\Users\asros\miniconda\envs\zeolites\lib\json\encoder.py:257](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:257), in JSONEncoder.iterencode(self, o, _one_shot)
    [252](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:252) else:
    [253](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:253)     _iterencode = _make_iterencode(
    [254](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:254)         markers, self.default, _encoder, self.indent, floatstr,
    [255](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:255)         self.key_separator, self.item_separator, self.sort_keys,
    [256](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:256)         self.skipkeys, _one_shot)
--> [257](file:///C:/Users/asros/miniconda/envs/zeolites/lib/json/encoder.py:257) return _iterencode(o, 0)

File [c:\Users\asros\miniconda\envs\zeolites\lib\site-packages\monty\json.py:394](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/json.py:394), in MontyEncoder.default(self, o)
    [391](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/json.py:391)     d = o.as_dict()
    [393](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/json.py:393) if "@module" not in d:
--> [394](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/json.py:394)     d["@module"] = str(o.__class__.__module__)
    [395](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/json.py:395) if "@class" not in d:
    [396](file:///C:/Users/asros/miniconda/envs/zeolites/lib/site-packages/monty/json.py:396)     d["@class"] = str(o.__class__.__name__)

TypeError: 'str' object does not support item assignment
```

### Proposed Solution

Modify the `.as_dict()` method?

### Alternatives

_No response_

Andrew-S-Rosen:
Closes #914.

```python
from monty.serialization import dumpfn
from emmet.core.symmetry import CrystalSystem

sym = CrystalSystem("Triclinic")

dumpfn(sym, "test.json")
```

This now works.

### DIFF
--- a/emmet-core/emmet/core/symmetry.py
+++ b/emmet-core/emmet/core/symmetry.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Any, Dict, Optional
 
 from pydantic import BaseModel, Field
@@ -6,12 +7,11 @@ from pymatgen.core.structure import Molecule
 from pymatgen.symmetry.analyzer import PointGroupAnalyzer, SpacegroupAnalyzer, spglib
 
 from emmet.core.settings import EmmetSettings
-from emmet.core.utils import ValueEnum
 
 SETTINGS = EmmetSettings()
 
 
-class CrystalSystem(ValueEnum):
+class CrystalSystem(Enum):
     """
     The crystal system of the lattice
     """
--- a/emmet-core/emmet/core/symmetry.py
+++ b/emmet-core/emmet/core/symmetry.py
@@ -1,4 +1,3 @@
-from enum import Enum
 from typing import Any, Dict, Optional
 
 from pydantic import BaseModel, Field
@@ -7,11 +6,12 @@ from pymatgen.core.structure import Molecule
 from pymatgen.symmetry.analyzer import PointGroupAnalyzer, SpacegroupAnalyzer, spglib
 
 from emmet.core.settings import EmmetSettings
+from emmet.core.utils import ValueEnum
 
 SETTINGS = EmmetSettings()
 
 
-class CrystalSystem(Enum):
+class CrystalSystem(ValueEnum):
     """
     The crystal system of the lattice
     """
--- a/emmet-core/emmet/core/utils.py
+++ b/emmet-core/emmet/core/utils.py
@@ -287,32 +287,7 @@ def jsanitize(obj, strict=False, allow_bson=False):
     return jsanitize(obj.as_dict(), strict=strict, allow_bson=allow_bson)
 
 
-class ValueEnum(Enum):
-    """
-    Enum that serializes to string as the value
-    """
-
-    def __str__(self):
-        return str(self.value)
-
-    def __eq__(self, obj: object) -> bool:
-        """Special Equals to enable converting strings back to the enum"""
-        if isinstance(obj, str):
-            return super().__eq__(self.__class__(obj))
-        elif isinstance(obj, self.__class__):
-            return super().__eq__(obj)
-        return False
-
-    def __hash__(self):
-        """Get a hash of the enum."""
-        return hash(str(self))
-
-    def as_dict(self):
-        """Create a serializable representation of the enum."""
-        return str(self.value)
-
-
-class DocEnum(ValueEnum):
+class DocEnum(Enum):
     """
     Enum with docstrings support
     from: https://stackoverflow.com/a/50473952
--- a/emmet-core/emmet/core/utils.py
+++ b/emmet-core/emmet/core/utils.py
@@ -5,11 +5,8 @@ from itertools import groupby
 from typing import Any, Dict, Iterator, List, Optional, Union
 
 import numpy as np
-
 from monty.json import MSONable
-
 from pydantic import BaseModel
-
 from pymatgen.analysis.graphs import MoleculeGraph
 from pymatgen.analysis.local_env import OpenBabelNN, metal_edge_extender
 from pymatgen.analysis.molecule_matcher import MoleculeMatcher
@@ -287,7 +284,28 @@ def jsanitize(obj, strict=False, allow_bson=False):
     return jsanitize(obj.as_dict(), strict=strict, allow_bson=allow_bson)
 
 
-class DocEnum(Enum):
+class ValueEnum(Enum):
+    """
+    Enum that serializes to string as the value
+    """
+
+    def __str__(self):
+        return str(self.value)
+
+    def __eq__(self, obj: object) -> bool:
+        """Special Equals to enable converting strings back to the enum"""
+        if isinstance(obj, str):
+            return super().__eq__(self.__class__(obj))
+        elif isinstance(obj, self.__class__):
+            return super().__eq__(obj)
+        return False
+
+    def __hash__(self):
+        """Get a hash of the enum."""
+        return hash(str(self))
+
+
+class DocEnum(ValueEnum):
     """
     Enum with docstrings support
     from: https://stackoverflow.com/a/50473952
--- a/emmet-core/setup.py
+++ b/emmet-core/setup.py
@@ -18,7 +18,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "pymatgen>=2023.10.11",
-        "monty>=2023.9.25",
+        "monty>=2024.2.2",
         "pydantic>=2.0",
         "pydantic-settings>=2.0",
         "pybtex~=0.24",
--- a/emmet-core/tests/test_utils.py
+++ b/emmet-core/tests/test_utils.py
@@ -1,10 +1,12 @@
 import datetime
 import json
+from pathlib import Path
 
 import numpy as np
 import pytest
 from bson.objectid import ObjectId
 from monty.json import MSONable
+from monty.serialization import dumpfn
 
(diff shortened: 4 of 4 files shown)


Andrew-S-Rosen:
Tagging @munrojm. Fixed it, I think! If you wouldn't mind minting a new version after this is merged, that would be a huge help.

Thank you @jmmshn for your suggestion. It worked like a charm! Not sure what I was doing wrong before.

jmmshn:
I will never end the crusade against custom `as_dict` / `from_dict`

esoteric-ephemera:
Hey I think this broke atomate2, see this [atomate2 issue](https://github.com/materialsproject/atomate2/issues/781) which I've reproduced with emmet>=0.78.0, and couldn't reproduce with emmet<=0.77.1

Andrew-S-Rosen:
It's probably better to solve the root cause of the problem in atomate2 (or monty, if relevant) rather than patch back in a custom `.as_dict()` method that also doesn't serialize to a dictionary but rather a string.

utf:
Maybe I'm missing something, but this PR changed the behaviour of these objects. That doesn't seem like an atomate2 issue as we were using these objects as intended.

esoteric-ephemera:
There's a `jsanitize` call in atomate2 with `strict=True` passed that would require `VaspObject` to have an `as_dict` method, which was removed here

utf:
The point is that we want to serialise these objects as strings. That's why we were using ValueEnum. 

I'd ask if you want to have enums that serialise as dicts then please make another object!

Andrew-S-Rosen:
I guess there are a few things to unpack here.

In emmet, and therefore in atomate2 and other packages, there are a few objects inheriting from `ValueEnum`s that break serialization. The example I mentioned in #944 is one example. When the `CrystalSystem` is dumped (e.g. deposited in a MongoDB), you can't reload it. That's because the custom `.as_dict()` method returning a `str` cannot reversibly generate back the parent `CrystalSystem`. This breaks roundtrip `dumpfn`/`loadfn` and is a problem.

The `ValueEnum` in emmet is, admittedly, advertised as a way to serialize the object to `str`. So, I can see the argument there for retaining the `.as_dict()` method to be a `str` return. However, the method is called `.as_dict()`. In my opinion, we should not have a method like `.as_dict()` returning something that is not a dictionary. This has downstream negative effects, as mentioned in the example above, and is arguably a hack.

I suppose the question to ask is there a better way to achieve the desired `str` serialization before without relying on such a hacky/fake `.as_dict()` method? I would argue the answer is yes --- inherit from `Enum` and take advantage of monty's `enum_values=True` [flag](https://github.com/materialsvirtuallab/monty/blob/4520972e78985394b3ef02d0b8ffb1873bc27a61/monty/json.py#L656).

utf:
I think this PR and issue #914 are misunderstanding what `ValueEnum` is for. There many objects in Python that cannot be serialised as json by themselves. For example, `str`, `int`, `float`, `list` etc. Instead, they have to be serialised as part of a dict.  The fact that this does not work

```python
from monty.serialization import dumpfn
from emmet.core.symmetry import CrystalSystem

sym = CrystalSystem("Triclinic")
dumpfn(sym, "test.json")
```
is not an issue in itself. This is equivalent to saying the following means that string serialisation is broken:

```python
dumpfn("Triclinic", "test.json")
```

The goal of `ValueEnum` is to define a set of string literals in the form of an enum. When we serialise these in a task document, we want to only store the enum value. E.g.

```python
{
    "crystal_system": "Triclinic"
}
```

rather than

```python
{
    "crystal_system": {
        "@module": "XYZ",
        "@class": "ValueEnum",
        "value": "Triclinic",
    }
}
```

The latter makes it convoluted to query the task document easily.

Perhaps there is a better way to achieve this and I agree that the `as_dict` labelling could be confusing. However, the point is this code has been in production for several years and works absolutely fine. The suggestion of enabling `enum_values=True` is not a good fix in this case, since this will mean it isn't possible to store ANY enums in their dict form.

To summarise, the issue raised in #914 is not a bug it is the intended behaviour as per the docstring and existing usage. This change has downstream impacts in atomate2. Please can we revert it.

PR #966, opened 2024-03-12
replace check for potcar hash with check for potcar summary stats

tsmathis:
The function `.get_potcar_hash()` has been removed from the `PotcarSingle` class in pymatgen. As a stop gap while we are migrating to new validation practices, we need to still be able to verify that the potcars used for a calculation match the pymatgen input set.

The proposed change from @esoteric-ephemera is to now check the potcar summary stats.


tsmathis:
tests for `emmet-core/tests/vasp/` need to be updated as well once the potcar checking is finalized

esoteric-ephemera:
We discussed this a bit and decided that for now, we'll maintain the legacy hash checking when the `summary_stats` kwarg isn't populated in a `TaskDoc` / `TaskDocument`. This field will be populated when parsing a VASP output directory, but is not currently populated in most (probably all) of the MP tasks.

Submitted a PR to this branch to add legacy checking as a fallback when the `summary_stats` kwarg isn't populated for any single potcar_spec.

PR #971, opened 2024-03-20
TaskDoc refactor

esoteric-ephemera:
## Summary
The main goal is removing atomate dependencies from emmet, and transitioning from using `emmet.core.vasp.task_valid.TaskDocumement` to `emmet.core.tasks.TaskDoc`. This will allow for eventual deprecation / aliasing of `TaskDocument`

## Completed:
Changes to `TaskDoc` and its fields are required since it is not directly compatible with `TaskDocument`:
- The fields of `TaskDocument` are deserialized, thus `TaskDocument.calcs_reversed[:].input` will be a dict whereas `TaskDocument.calcs_reversed[:].input` is a `CalculationInput` object. Calling `.model_dump()` doesn't help since this isn't recursive (e.g., the structure in CalculationInput will still be serialized)
- Added a `BaseCalculationModel` to `emmet.core.calculation` which just wraps the pydantic `BaseModel` with a `.get` method to emulate dict key access (needed in the ValidationDoc)
- Added a few fields to add missing attrs from `TaskDocument`: `calc_type`, `run_type`, and a post init update for `task_type`
- Because the `TaskDoc` accepts extra model input, and the way that pydantic looks for non-field attrs in the class, there are new private fields defined on `TaskDoc`. pydantic first looks for attrs in the model extra field, but then raises an `AttributeError` if it can't find it, without looking to see if `hasattr(class,attr)`. There's a longer explanation in a comment near `_calc_type`

Will be part of a separate PR:
~## To Do:~
~- Remove atomate dependence from emmet-cli: `VaspCalcDb` --> generic maggma store~
~- Improve credential handling in emmet-cli settings~


mkhorton:
If there are breaking changes to `TaskDoc`, eg for current users of atomate2, could a migration script be provided?

tsmathis:
merging as rc

PR #972, opened 2024-03-25
Fix potcar checking

esoteric-ephemera:
In a previous PR, I set the "titel" field of the POTCAR spec to `pymatgen.io.vasp.inputs.PotcarSingle.TITEL`. *However*, there is an almost-always identical `header` attr on `PotcarSingle` which should be used in the spec instead

On the builders side, this is needed for `emmet.builders.utils.get_potcar_stats` method, which is used to validate calculations. The `header` attr of `PotcarSingle` is now used in this function. I've also added the option to load a dict of pre-defined POTCAR specs with this function, and added such a dict for MP calcs

Added tests (two that don't rely on a POTCAR library being established) for `get_potcar_stats`

For backwards compatibility reasons, I don't think we can change the `titel` attr of `emmet.core.vasp.calculation.PotcarSingle` to be `header`, but maybe we can simply alias it for now and eventually deprecate it. This notation is confusing / not correct


tsmathis:
Looks good on my end. Glad we’re getting the checks ironed out.

PR #1011, opened 2024-05-09
Change default values in document model for TaskDocument.output.stress/forces

tsmathis:
Pydantic will throw validation errors for older calculations where stress and forces were not calculated. An empty list as the default field causes pydantic to check the empty list for Matrix3D or Vector3D entries, which don't exist.


tsmathis:
tests in core for migration_graph not passing due to pmg issue [here](https://github.com/materialsvirtuallab/pymatgen-analysis-diffusion/pull/389). Will go ahead anyway since this is just for an rc

PR #1018, opened 2024-05-30
Refactor Taskdoc.orig_inputs.potcar and bug fix

esoteric-ephemera:
As part of the effort in refining the TaskDoc scheme with @tsmathis, we're restricting the `TaskDoc.orig_inputs.potcar` field to be an `emmet.core.tasks.Potcar` object. The user can still input a serialized pymatgen `Potcar` object to `TaskDoc.orig_inputs.potcar` (no breaking changes), however the `OrigInputs` class will change this to the emmet-core `Potcar` representation.

Also a small bug fix for the builder potcar library caching (wrong kwarg).


munrojm:
I am highly in favor of this.

Issue #960, opened 2024-03-02
Bug: Wrong `run_type` report on r2SCAN run


Andrew-S-Rosen:
### Problem

```python
from emmet.core.tasks import TaskDoc
task_doc = TaskDoc.from_directory(".")
print(task_doc.entry.parameters["run_type"]) # prints GGA even though it's an r2SCAN run
```

We can strip back a few layers to find the root cause of the problem:

```python
from pymatgen.io.vasp.outputs import Vasprun
from emmet.core.vasp.calculation import CalculationInput
from emmet.core.vasp.calc_types import run_type

vr = Vasprun("vasprun.xml.gz")
doc = CalculationInput.from_vasprun(vr)
print(run_type(doc.parameters)) # prints GGA
```

So, the source of our problems here is an incorrect parse on the `vasprun.xml.gz` parameters. Going one step deeper, we can see

```python
vr.parameters["METAGGA"] # raises a KeyError
```

So, this is actually a Pymatgen issue.... reporting upstream.

[vasprun.xml.gz](https://github.com/materialsproject/emmet/files/14470586/vasprun.xml.gz)
[OUTCAR.gz](https://github.com/materialsproject/emmet/files/14470587/OUTCAR.gz)
[POSCAR.gz](https://github.com/materialsproject/emmet/files/14470588/POSCAR.gz)
[INCAR.gz](https://github.com/materialsproject/emmet/files/14470589/INCAR.gz)
[CONTCAR.gz](https://github.com/materialsproject/emmet/files/14470590/CONTCAR.gz)
Feel free to email me for the POTCAR.

### Proposed Solution

TBD.

Andrew-S-Rosen:
@esoteric-ephemera, any idea on this one?

Andrew-S-Rosen:
Closing this as being a pymatgen issue.

Andrew-S-Rosen:
I'm reopening this because Pymatgen is technically doing what it's told. The `METAGGA` (and other fields, like `ALGO`) are missing entirely from the `<parameters>` block of the `vasprun.xml` file even though they are present in the `<incar>` section. Pymatgen then doesn't report "METAGGA" in its `parameters` attribute of the `vasprun.xml` file because it doesn't exist. This can be problematic when generating an entry because the `run_type` will be wrong, which then also means that using things like the MP compatibility schemes will be wrong. I'm not sure the best way to address this one...

esoteric-ephemera:
[...]

Issue #1021, opened 2024-06-02
[Help Wanted]: Can not parse trajectory of VASP optimization directory


wangzyphysics:
### Problem

when trying to parse the structure optimization direcotry of VASP, i encounter this error:

```bash
> python run.py 
Traceback (most recent call last):
  File "/home/xxx/workplace/mp-1198074-prim/run.py", line 5, in <module>
    d = TaskDoc.from_directory("./", store_trajectory=StoreTrajectoryOption.FULL)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/softs/miniconda3/envs/emmet/lib/python3.12/site-packages/emmet/core/tasks.py", line 481, in from_directory
    analysis = AnalysisDoc.from_vasp_calc_docs(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/softs/miniconda3/envs/emmet/lib/python3.12/site-packages/emmet/core/tasks.py", line 305, in from_vasp_calc_docs
    max_force = _get_max_force(final_calc)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/softs/miniconda3/envs/emmet/lib/python3.12/site-packages/emmet/core/tasks.py", line 851, in _get_max_force
    forces: Optional[Union[np.ndarray, List]] = calc_doc.output.ionic_steps[-1].forces
                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
TypeError: 'NoneType' object is not subscriptable
```

and the context of `run.py` is :

```python
from emmet.core.tasks import TaskDoc
from emmet.core.vasp.calculation import StoreTrajectoryOption


d = TaskDoc.from_directory("./", store_trajectory=StoreTrajectoryOption.FULL)
print(d)
```

How should i fix this? 

### Proposed Solution

None

tsmathis:
Hi @wangzyphysics, the traceback from your error implies that your calculation is missing the `output.ionic_steps` field, which might be hard to diagnose on our end. 

But for starters, what version of `emmet-core` are you using?

wangzyphysics:
> Hi @wangzyphysics, the traceback from your error implies that the calculation doc that is being generated from your calculation directory is missing the `output.ionic_steps` field, which might be hard to diagnose on our end.
> 
[...]

PR #1026, opened 2024-06-06
Update validation doc to use TaskDoc + add nelements and symmetry_number as fields in document model

tsmathis:
updates ValidationDoc to use the new atomate 2 TaskDoc

Also added nelements and symmetry_number as optional fields for use when grouping documents together for object storage, i.e., s3


tsmathis:
@esoteric-ephemera , would you do anything differently here?

esoteric-ephemera:
This looks good! I would only change `task_doc: TaskDocument | TaskDoc` so that it's backwards compatible for now. The line about model dumping in `ValidationDoc`:
```
calcs_reversed = [
            calc if not hasattr(calc, "model_dump") else calc.model_dump()
            for calc in task_doc.calcs_reversed
]
```
is there to support `TaskDocument` (`calcs_reversed` is a dict) and `TaskDoc` (`calcs_reversed` is a pydantic model object).

Obvs when we're ready to fully deprecate `TaskDocument`, we can remove that glue code and type suggestion.

tsmathis:
Correct me if I'm wrong, but isn't TaskDocument not compatible with the dot notation access that is used for TaskDoc's props now?

See the changes I had to make here: f5f52fb

esoteric-ephemera:
You're right about the dot notation, but we can support both if you change _magmom_check to depend only on calcs_reversed, since those lines I pointed out ensure that's always deserialized. The full task doc isn't needed for the magmom check, just calcs reversed

Also `TaskDoc.output.structure` should be decorated with magnetic moments, so returning something like
`any(site.properties.get("magmom",0.) > eles_max_vals.get(site.label, 0.) for site in structure)`
should work in place of the current very long check. (The structure needs to be serialized for that to work.)

Happy to add these changes if that's easier

tsmathis:
Yeah feel free to add that in if you know what to change off the top of your head. I'd have to trial and error my way through it. 

I'm running the validation builder over the full set of task docs to see if there are other gotchas as well

tsmathis:
@munrojm, any comment/issue here? I've run the validation builder with the legacy validation settings on the 300k+ task docs under the `mp_production_old` tag and everything ran smoothly there. 

Working through the remainder of the newer production tags now, but I don't think there should be issues with the newer calcs.

esoteric-ephemera:
@munrojm also want to note that this PR closes issues #960 and #1021

munrojm:
No further comments from me, this looks good.

PR #1033, opened 2024-06-12
Ensure run, calc, and task types are always refreshed in TaskDoc

esoteric-ephemera:
Simple changes to `TaskDoc` structure: during post init, the `run_type`, `task_type` and `calc_type` (union of the previous two) are always updated to ensure that these are consistent with subsequent bug fixes. Previously, these were only updated if `task_type` was `None`.

This change is motivated by [#960](https://github.com/materialsproject/emmet/issues/960), where the `run_type` was incorrectly determined for any non-GGA run and some discussions with @mkhorton. By re-initializing a `TaskDoc` from a dict, these changes would correctly populate the three fields. 

The source of the bug in 960 is separately fixed in [#1026](https://github.com/materialsproject/emmet/pull/1026). This change would also ensure that 1026 also patches fields in legacy TaskDocs

### DIFF
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -370,6 +370,15 @@ class TaskDoc(StructureMetadata, extra="allow"):
         None, description="The type of calculation."
     )
 
+    run_type: Optional[RunType] = Field(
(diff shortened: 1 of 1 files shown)


munrojm:
Looks good to me!

PR #1040, opened 2024-06-24
Standardize TaskDoc input fields

esoteric-ephemera:
Two primary changes:
- Ensure that `TaskDoc.input`, `TaskDoc.orig_inputs`, and `TaskDoc.calcs_reversed.*.input` all inherit from `emmet.core.vasp.calculation.CalculationInput`. The top-level `input` and `orig_inputs` retain their extra fields
- Change prioritization of how task type is determined from inputs (`calcs_reversed[0].input` first, top-level `input` second, and lastly `orig_inputs`)
 
This should close [#1021](https://github.com/materialsproject/emmet/issues/1021): the task type can be determined without running with custodian (the `orig_inputs` field is `None)

### DIFF
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -21,7 +21,7 @@ from emmet.core.vasp.calc_types import (
     task_type,
 )
 from emmet.core.vasp.calculation import (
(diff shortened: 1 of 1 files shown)


tsmathis:
Nice, works as expected!

PR #1060, opened 2024-07-23
Remove atomate dependence from emmet-cli / charge density analysis improvements

esoteric-ephemera:
Fully removes the atomate dependence from emmet-cli. `VaspDrone` has been replaced with `TaskDoc.from_directory`, and the `VaspCalcDb` has been replaced with a `TaskStore` object that repurposes only its core internals.

Also adds ability to parse existing bader / ddec6 charge density analysis files during `TaskDoc` creation / CLI parsing. Should save some time if these are precomputed and stored.


esoteric-ephemera:
Superseded by #1229

PR #1073, opened 2024-08-16
Configure query on request

yang-ruoxi:
# summary

- Added a method in `GlobalHeaderProcessor` to translate the header to a license query 
- Added `None` to as allowed license field in `LicenseQuery` 
- Use the the `configure_query_on_request` method in materials/core and materials/summary resources 

Note that before we actually put in the actual ACLs for our users, the api and website will behave the same. 
After we update the data, we will prompt the users to accept terms again, and the ACLs will be automatically assigned and this header thing will take effect.


tschaume:
@yang-ruoxi the tests are running again. The latest error is from mypy:

```
Run python3.11 -m mypy --namespace-package emmet-api/emmet
emmet-api/emmet/api/core/global_header.py:29: error: Unexpected keyword argument "license" for "query" of "QueryOperator"  [call-arg]
/home/runner/.local/lib/python3.11/site-packages/maggma/api/query_operator/core.py:15: note: "query" of "QueryOperator" defined here
```

PR #1076, opened 2024-08-20
Add convenience util func for generating robocrys descriptions

tsmathis:
@esoteric-ephemera, any extra additions when you get a chance.

Likely need some flexibility with kwargs, or any defaults you think are sane.


tsmathis:
@esoteric-ephemera, can you take a look at this when you have a chance? Not sure if this still applies with the recent robocrys updates

esoteric-ephemera:
Yes! Have some specific thoughts right now having run this a bunch: there are a few common exceptions that get thrown by robocrys that I don't think we're handling:
- 'Symmetry detection failed for structure with formula < formula >. Try setting symprec=0.01 to a different value.'
- "2D components don't all have the same orientation."
- 'No Voronoi neighbors found for site - try increasing cutoff'
- 'No unicode character exists for subscript period.'

I think we can maybe treat the first one by adjusting `symprec` thru the kwargs, the last one (missing unicode character) probably needs a fix on the robocrys side

The Voronoi exception is from pymatgen, might be too buried to adjust through robocrys kwargs but I'll look into that

Will also add some defaults logic - we probably want some additions to the builder too to reduce load time in mineral matcher

tsmathis:
All sounds good

From the builder side the plan was just to pass in an already instantiated mineral matcher to the `from_structure` method, which can be reused in further calls to `from_structure`.

Just instance like `mineral_matcher = MineralMatcher()`, should there be anything different there?

esoteric-ephemera:
@tsmathis sorry for the delay, adding some extra logic for handling some common exceptions - think this is in good shape

tsmathis:
nice, thanks!

tsmathis:
couple things, since robocrys is optional how do you want to handle the `MineralMatcher` imports? In the scope of this the `MineralMatcher` is only imported for type hinting

esoteric-ephemera:
> one thing, since robocrys is optional for emmet how do you want to handle the MineralMatcher imports? In the scope of this the MineralMatcher is only imported for type hinting

Since we're using the legacy type hints, I think your solution (just removing the type hint for `mineral_matcher`) is OK!

PR #1010, opened 2024-05-07
A builder for classical md workflows

orionarcher:
This PR implements `ElectrolyteBuilder` and `SolventBenchmarkBuilder` classes for analysis of trajectories generated by atomate2.openmm.

The core Builder classes are supported by a variety of utility functions that interact with `MDAnalysis`, `SolvationAnalysis`, and `TransportAnalysis`. In addition to the usual `get_items`, `process_items`, `update_targets` workflow, users can also manually create an `MDAnalysis.Universe` directly from the taskdoc with `ElectrolyteBuilder.instantiate_universe`. This makes manual analysis of a small number of systems much easier.

Everything is reasonably well tested and once CI passes this will be ready for review.

PR #1087, opened 2024-09-05
Refactor and small updates to OpenMM and OpenFF task documents

orionarcher:
This PR implements changes to the core OpenMM and OpenFF task docs to support changes to the `atomate2` workflows downstream. It is intended to split off the functionality changes in `emmet-core` implemented in PR #1010, which was becoming too large and burdensome.
- [x] changes the directory structure
- [x] change the names and type signatures of some task_doc elements

I cleaned up the commit history so that it's easier to inspect. @tsmathis.


tsmathis:
nice, thanks @orionarcher

just to confirm this one will get your downstream work in atomate2 moving?

orionarcher:
Yep, just tested this with the atomate2 code and it's all working locally.

emmet-core/emmet/core/openff/core.py
```diff
@@ -0,0 +1,9 @@
+from emmet.core.openff import MoleculeSpec
+from emmet.core.base import EmmetBaseModel
+
+
+class ClassicalMDDoc(EmmetBaseModel):
```
tsmathis, line 5: there is a class definition for SolvationDoc that is commented out in "emmet-core/emmet/core/openff/solvation.py" (line 30), outside of that I don't see other references to this ClassicalMDDoc model.

Should this be used? If not, should it be removed?

emmet-core/emmet/core/openmm/calculations.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    @classmethod
+    def from_calcs_reversed(
+        cls,
+        calcs_reversed: List[Calculation],
+        job_uuid: Optional[str] = None,
+        flow_uuid: Optional[str] = None,
+    ) -> "CalculationsDoc":
+        calcs = calcs_reversed[::-1]
```
tsmathis, line 58: Is there a reason to un-reverse the calcs_reversed here? 

From this I understand that the entries in each of the arrays for the document fields will correspond to the order in which the calcs were completed, but would that be confusing given the calcs_reversed are in the opposite order? 

I see this is still taking the last calc (after un-reversing) for the completed_at field.

emmet-core/setup.py
```diff
@@ -27,8 +27,7 @@
     },
     include_package_data=True,
     install_requires=[
-        "numpy<2",
-        "pymatgen",
+        "pymatgen==2024.4.13",
```
tsmathis, line 30: do we need to pin pmg?

emmet-core/setup.py
```diff
@@ -40,10 +39,9 @@
             "matcalc>=0.0.4",
             "seekpath>=2.0.1",
             "robocrys>=0.2.8",
-            "pymatgen-analysis-defects>=2024.7.18",
-            "pymatgen-analysis-diffusion>=2024.7.15",
-            "pymatgen-analysis-alloys>=0.0.6",
-            "solvation-analysis>=0.4.1",
+            "pymatgen-analysis-diffusion>=2023.8.15",
```
tsmathis, line 42: and do we need to downgrade the analysis libraries?

emmet-core/emmet/core/openmm/calculations.py
orionarcher: I want the lists in `CalculationsDoc` to follow a sequential order rather than reversed because it's more intuitive. The "reversal" of `calcs_reversed` is a bit less intuitive IMO (as evidenced by it being called `calcs_reversed` rather than `calcs`).

I don't want to have `task_names_reversed`, `calc_types_reversed`, etc.

emmet-core/emmet/core/openff/benchmarking.py
```diff
@@ -0,0 +1,99 @@
...
+    ) -> "SolventBenchmarkingDoc":
+        if temperature is not None:
+            dielectric = DielectricConstant(
+                u.atoms, temperature=temperature, make_whole=False
+            )
+            dielectric_run_kwargs = dielectric_run_kwargs or {}
+            dielectric.run(**dielectric_run_kwargs)
+            eps = dielectric.results.eps_mean
+        else:
+            warnings.warn(
```
tsmathis, line 65: for these warnings, when we're going to run the builders in a distributed cluster how are we meant to suppress these?

Just with this on the builder side?
```
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    SolventBenchmarkingDoc.from_universe(...)
    etc.
```

tsmathis: same thought on the other `warnings.warn`s

emmet-core/emmet/core/openff/core.py
orionarcher: This is a bit aspirational so I can remove it for now. The intent is for this class to eventually contain the results of all the different builders.

emmet-core/emmet/core/openmm/calculations.py
tsmathis: Yep, agreed,  makes sense. But a small doc string might be helpful since all the other workflows, doc models, and builders do follow the calcs_reversed pattern

emmet-core/setup.py
orionarcher: This is a mistake introduced in the rebase, my apologies will fix

emmet-core/setup.py
orionarcher: Will remove

emmet-core/emmet/core/openff/benchmarking.py
orionarcher: Very good point, this isn't something I was considering. I am happy to implement whatever solution is best/standard. If that's the solution you described above I can add it in now.

tsmathis: we have the filter warnings function in other builders, so I would deem that as okay if having easy access to these warnings at the document model is important to you for testing, etc.

Could also probably do something will the logger for the builder to move these messages over to the builder side, but that isn't entirely necessary at this point imo

emmet-core/emmet/core/openff/core.py
tsmathis: okay, good to know. can be re-implemented in the future when you have that fleshed out

emmet-core/emmet/core/openff/benchmarking.py
orionarcher: There are some warning's in PR #1088 that I'll do that for but I'll just ax this one. It's not really user facing so the temperature should always be available.

emmet-core/emmet/core/openff/benchmarking.py
```diff
@@ -0,0 +1,96 @@
...
+                temp_avg=temperature,
+                linear_fit_window=(start, stop),
+            )
+            viscosity_run_kwargs = viscosity_run_kwargs or {}
+            viscosity_helfand.run(**viscosity_run_kwargs)
+            viscosity_function_values = viscosity_helfand.results.timeseries.tolist()
+            viscosity = viscosity_helfand.results.viscosity
+
+        else:
+            warnings.warn(
```
tsmathis, line 80: if you're removing the warnings, here's the last one

orionarcher: I think I got them all.

emmet-core/emmet/core/openmm/calculations.py
orionarcher: Added a docstring to clarify.

tsmathis:
any further changes you want in @orionarcher?

tsmathis:
@tschaume, some updates to reqs files were needed to get this through, namely a pin on `dgl` for the ml extras. But this is likely since I was compiling the reqs locally.

But it does seem like dgl does not have a ubuntu-compatible version >2.1 on pypi (failed actions here: [failed run](https://github.com/materialsproject/emmet/actions/runs/10728003240/job/29751596953#step:5:83)

so a re-run of the deps upgrade will be needed I think after this is in

tschaume:
@tsmathis ok, thanks for the heads-up! I can rerun the dependencies and take a closer look after this is merged.

orionarcher:
This is good to merge. If there are any other changes/updates I can include them in a future smaller PR.

tsmathis:
> ~This is good to merge. If there are any other changes/updates I can include them in a future smaller PR.~
> 
> Scratch that, found some weird interactions with jobflow-remote that I am working on debugging.

All good, either ping us here or on slack when you're ready

orionarcher:
@tsmathis, fixed. This should be good to merge when tests pass.

tsmathis:
👍

PR #1094, opened 2024-09-10
Update typing in EmmetSettings

tsmathis:
pydantic plus python 3.10 did not like `dict` type annotations, switched to `Dict`


tsmathis:
superseded by #1095

PR #1095, opened 2024-09-10
Generate enums in Github Actions

tschaume:
@esoteric-ephemera @tsmathis Regenerating the enums removes the `_missing_` classmethod for the VASP calc types. Is that being used/needed anywhere?


tsmathis:
I'm not seeing the use of `__missing__` anywhere in emmet

tsmathis:
I'll also close #1094 in favor of this

tsmathis:
> I'm not seeing the use of `__missing__` anywhere in emmet

sorry, that's `_missing_` actually isn't it. But same thing, non-existent as far as I can tell

esoteric-ephemera:
@tschaume, `_missing_` is needed for case insensitivity, see [this issue](https://github.com/materialsproject/api/issues/869) that @munrojm fixed a while ago. Without `_missing_`, pydantic throws validation errors (checked for that specific MPID in the issue, `mp-2048877`), adding it back resolves them

I'd like to take a crack at having the action generate a yaml file, and instantiating the enums from yaml. I'll make a separate PR for that

PR #1116, opened 2024-09-25
add completed_at field (+run_stats) to TaskDoc as root fields

tsmathis:
@esoteric-ephemera, this should be enough to get `completed_at` in as a root level field, right?

There were already kwargs of `completed_at=calcs_reversed[0].completed_at,` in the `.from_structure` calls, but the field isn't showing up currently

### DIFF
--- a/emmet-builders/tests/test_utils.py
+++ b/emmet-builders/tests/test_utils.py
@@ -68,7 +68,9 @@ def test_get_potcar_stats(method: str, tmp_path):
     try:
         potcar_stats = get_potcar_stats(method=method)
     except Exception as exc:
-        if "No POTCAR for" in str(exc):
+        if any(
+            exc_str in str(exc) for exc_str in ("Set PMG_VASP_PSP_DIR", "No POTCAR for")
+        ):
             # No Potcar library available, skip test
             return
         else:
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -434,6 +434,11 @@ class TaskDoc(StructureMetadata, extra="allow"):
         description="Identifier for this calculation; should provide rough information about the calculation origin and purpose.",
     )
 
+    run_stats: Optional[RunStatistics] = Field(
+        None,
+        description="Summary of runtime statistics for each calculation in this task",
+    )
+
     # Note that private fields are needed because TaskDoc permits extra info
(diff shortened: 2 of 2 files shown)


esoteric-ephemera:
@tsmathis that should be good - also worth noting that `run_stats` and `include_structure` are two other optional fields that are automatically populated by `TaskDoc.{from_directory, from_vasprun}`. We should probably make these default top-levels too right?

Also tweaked one of the builder tests to better account for when no POTCAR library is preset (like CI)

tsmathis:
I can't really find anywhere that `include_structure` actually has an effect?

And the `run_stats` would be okay to include I guess since they were in `TaskDocument`. Will be just another field that we'll have to drop when parsing though.

There is also differing behavior for `from_directory` and `from_vasprun` for the run stats field, should they be the same?

esoteric-ephemera:
I'm not sure what the purpose of `include_structure` is - it's not used anywhere in emmet or atomate2 and it's just a bool. I'd be tempted to change it to a property of `TaskDoc` that returns whether `TaskDoc.structure` is `None`

`from_directory` and `from_vasprun` have slightly different outputs - the `from_vasprun` method doesn't populate as many fields because it doesn't have access to things like volumetric objects. Their signatures are the same tho

esoteric-ephemera:
Hey @utf, do you recall what the purpose of `TaskDoc.include_structure` is? We'd like to remove it from the `TaskDoc.from_directory` and `TaskDoc.from_vasprun` methods, since it's not defined in the base schema of `TaskDoc`.

utf:
IIRC it was an option as to whether the structure was stored alongside the metadata. It was originally part of StuctureMetadata.from_structure. I don't remember it being a field though.

tsmathis:
That was the impression we got based on the name, but we haven't found any place in either emmet or atomate2 where `include_structure` has an effect for TaskDoc. 

Is this behavior you'd like to see functioning for TaskDoc?

utf:
As long as the structure is always stored in the task doc that's fine. Sounds like it's safe to remove.

tsmathis:
Okay, sounds good @utf, thanks for the input. 

@esoteric-ephemera, lets follow up separately on the `include_structure` behavior and make sure no regressions are introduced.

That aside, going to merge this in unless there is anything else you'd like to see in these current changes?

PR #1118, opened 2024-09-28
Improve error handling in `SymmetryData`

mkhorton:
Hi everyone, just a small PR:

The `SymmetryData` model currently has `optional` fields, and therefore can be constructed even if symmetry information is unknown, e.g. if spglib fails.

This PR improves error handling in `.from_structure` to gracefully handle cases where spglib cannot determine the symmetry of a structure. A common use case where this might be encountered is if an otherwise valid `StructureMeta` (or a subclass of `StructureMeta`) is instantiated, but with symmetry information unknown.

The PR also improves handling of `angle_tolerance` to ensure that this is picked up from `EmmetSettings` instead of assuming the default value, and records this alongside `symprec` in the document model. This is also an optional field so as not to break backwards compatibility.

It might be better to make these fields non-optional, but that would be a breaking change and I don't want to suggest it in this PR.

### DIFF
--- a/emmet-core/emmet/core/symmetry.py
+++ b/emmet-core/emmet/core/symmetry.py
@@ -3,7 +3,12 @@ from typing import Any, Dict, Optional
 from pydantic import BaseModel, Field
 from pymatgen.core import Structure
 from pymatgen.core.structure import Molecule
-from pymatgen.symmetry.analyzer import PointGroupAnalyzer, SpacegroupAnalyzer, spglib, SymmetryUndeterminedError
+from pymatgen.symmetry.analyzer import (
+    PointGroupAnalyzer,
+    SpacegroupAnalyzer,
+    spglib,
+    SymmetryUndeterminedError,
+)
 
 from emmet.core.settings import EmmetSettings
 from emmet.core.utils import ValueEnum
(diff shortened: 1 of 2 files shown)


tsmathis:
Awesome, thanks @mkhorton. Is there a reason for the minimum pymatgen version? And if you can fix the linting that would be great.

And apologies for the failing workflows for the tests., we haven't configured the enum generation workflow to be compatible with forks. I've ran the tests locally with your patch and everything looks good.

mkhorton:
No worries about the tests from my side, I did test this myself also.

For the minimum pymatgen version, this is because `SymmetryUndeterminedError` was only added back in June. Previously this kind of error wasn't handled well in pymatgen either, for example `Structure.get_spacegroup_info()` would raise an odd `KeyError` if symmetry wasn't determined because the symmetry dataset dict would be `None`. But in either case, the minimum pymatgen version set is already an earlier than the versions pinned in the `requirements/` files, so it should be ok.

tsmathis:
Sounds good, just wanted to be sure there's a reason if we're asked in the future.

mkhorton:
Just ran through `black`, hopefully that fixes the linter issues. I don't actually have a dev environment set up for `emmet` right now otherwise, this is a bit of a hot fix :)

tsmathis:
All good, pre-commit complains a lot.

You happy with this? Can merge whenever you're ready

mkhorton:
I'm happy with this, thanks Tyler!

PR #1123, opened 2024-10-04
Add NEB schemas

esoteric-ephemera:
Adding support for parsing NEB calculations and eventually ApproxNEB calculations. These are still VASP calculations, but have a directory structure which is not parseable by TaskDoc (at least meaningfully). The structure is:
./{INCAR, KPOINTS, POTCAR} 
   ./00/POSCAR
   ./01/{vasprun.xml, OSZICAR, OUTCAR, CHGCAR, ... }
   ./02/{vasprun.xml, OSZICAR, OUTCAR, CHGCAR, ... }
   ...
   ./0N/POSCAR

Approx / NEB calculations are needed for battery cathode and alloy diffusivity calculations. These schemas will be needed for atomate2 workflows. Once Approx / NEB flows are added to atomate2, we should have full parity with atomate. This will also make it easier to host data from these calculations

Supports parsing NEB calculations (only relaxes images) and NEB + endpoints (three total calculations, two relaxes + one NEB)

Possibly worth adding run / task / calc types? Since this requires a separate doc model, maybe we should avoid this?


esoteric-ephemera:
@tschaume / @tsmathis : the document models / updates for NEB should be more or less final.

Since these changes are isolated to emmet-core's NEB functionality (completely new) + MigrationGraphDoc (revised by @hmlli to fix bugs), can I merge this in?

tschaume:
Thanks @esoteric-ephemera! I think it's ok to merge and put out a release candidate if needed.

PR #1124, opened 2024-10-07
Fix for MaterialsBuilder when `TaskDoc.tags` is `None`

mkhorton:
The schema allows `tags` to be `None` but the materials builder will fail if tags are not specified.

I have added a comment to explain what the `commercial_license` variable means. I think this is correct?

### DIFF
--- a/emmet-builders/emmet/builders/vasp/materials.py
+++ b/emmet-builders/emmet/builders/vasp/materials.py
@@ -238,9 +238,11 @@ class MaterialsBuilder(Builder):
         grouped_tasks = self.filter_and_group_tasks(tasks, task_transformations)
         materials = []
         for group in grouped_tasks:
+            # commercial_license = True means that the default CC-BY license is applied
+            # commercial_license = True means that a CC-BY-NC license is applied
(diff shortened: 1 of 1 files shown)


emmet-builders/emmet/builders/vasp/materials.py
```diff
@@ -238,9 +238,11 @@ def process_item(self, items: List[Dict]) -> List[Dict]:
         grouped_tasks = self.filter_and_group_tasks(tasks, task_transformations)
         materials = []
         for group in grouped_tasks:
+            # commercial_license == True means that the default CC-BY license is applied
+            # commercial_license == False means that a CC-BY-NC license is applied
             commercial_license = True
             for task_doc in group:
-                if set(task_doc.tags).intersection(
+                if task_dog.tags and set(task_doc.tags).intersection(
```
tsmathis, line 245: typo here, task_dog?

mkhorton: Thanks Tyler 🤦🏻‍♂️ Monday mornings

PR #1151, opened 2024-11-29
Make species hash and coord hash optional in `MoleculeMetadata`

### LINKED ISSUE #1149: [Feature Request]: Make OpenBabel optional for molecule metadata
### Problem

In emmet-core<=0.84.2, it was possible to make `MoleculeMetadata` without relying on OpenBabel as a requirement. However, in emmet-core==0.84.3rc4, OpenBabel is now implicitly a requirement because it is used in making the [species hash](https://github.com/materialsproject/emmet/blob/a705a6148b97803d30899d9c7ac3ca51c7e27136/emmet-core/emmet/core/structure.py#L316).

OpenBabel is a very annoying dependency to require because it cannot be pip installed. It would be very nice to maintain the old behavior and support the creation of `MoleculeMetadata` without OpenBabel as a dependency. I propose that we generate a different hash if OpenBabel is not around (nor not make one at all).

```python
from pymatgen.core import Molecule
from emmet.core.structure import MoleculeMetadata

mol = Molecule(['H','H'],[[0,0,0],[1,1,1]])
MoleculeMetadata.from_molecule(mol)
```

Andrew-S-Rosen:
This PR closes #1149 by making the species hash and coord hash fields optional in `MoleculeMetadata`, only generating them if OpenBabel is installed. In return, it is possible to generate `MoleculeMetadata` without relying on OpenBabel.

An alternate approach was given in https://github.com/materialsproject/emmet/pull/1150 where I used `JMolNN` as a fallbak option in place of `OpenBabelNN`, but I feel like the current PR is more suitable because it might be difficult to debug a scenario where a change in the environment's dependencies yields a different hash.

To reiterate what I mention in the issue report, the motivation behind this PR is that OpenBabel is a really annoying dependency to require since it is not on PyPI. This means any workflow infrastructure wanting to rely on `MoleculeMetadata` must rely on the user to install OpenBabel from source or Conda unless this PR is merged.

### DIFF
--- a/emmet-core/emmet/core/structure.py
+++ b/emmet-core/emmet/core/structure.py
@@ -20,6 +20,7 @@ try:
 except Exception:
     openbabel = None
 
+
 class StructureMetadata(EmmetBaseModel):
     """Mix-in class for structure metadata."""
 
--- a/emmet-core/emmet/core/structure.py
+++ b/emmet-core/emmet/core/structure.py
@@ -15,6 +15,10 @@ from emmet.core.symmetry import PointGroupData, SymmetryData
 T = TypeVar("T", bound="StructureMetadata")
 S = TypeVar("S", bound="MoleculeMetadata")
 
+try:
+    from openbabel import openbabel
+except Exception:
+    openbabel = None
 
 class StructureMetadata(EmmetBaseModel):
     """Mix-in class for structure metadata."""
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
Also, if approved, would it be possible to get a version release (or pre-release) after this PR? 🙏

tsmathis:
Thanks @Andrew-S-Rosen, this seems reasonable. Happy to merge this in if you're satisfied?



Will also get a release out ASAP as well

Andrew-S-Rosen:
Thank you, @tsmathis!!!

tsmathis:
@Andrew-S-Rosen, we're working through a deployment for a full emmet release, but for now try using the 0.84.3rc6 pre-release and let us know if anything isn't working on your end!

Andrew-S-Rosen:
Thanks! Seems to work!

PR #1184, opened 2025-01-28
Read POTIM value as time step

jmmshn:
The time-step for the trajectory object can be set from. 

Not sure how VASP focused this function is suppose to be. but the other ones in the file does assume VASP so I assume this is OK.


esoteric-ephemera:
@jmmshn: closing old/stale PRs, will incorporate this change with a check for an appropriate IBRION value (0 for MD) in #1230

PR #1185, opened 2025-01-28
Resolve mypy issues in emmet-api

esoteric-ephemera:
Fixing mypy issues in `emmet-api` related to the fact that pymatgen's `Composition` can take `Element`, `Species`, and `DummySpecies` as keys, and only the former two have a `name` attr that is a `str`

### DIFF
--- a/emmet-api/emmet/api/routes/materials/materials/query_operators.py
+++ b/emmet-api/emmet/api/routes/materials/materials/query_operators.py
@@ -409,12 +409,16 @@ class FormulaAutoCompleteQuery(QueryOperator):
             comp_red = comp.reduced_composition.items()
 
             for i, j in comp_red:
+                # The keys of pymatgen's Composition can be Element, Species, or DummySpecies
+                # Element and Species both have a name attr, DummySpecies doesn't by default
+                # This bothers mypy a lot, so we placate it here - all three have __str__ methods:
+                spec_name = str(getattr(i, "name", None) or i)
                 if j != 1:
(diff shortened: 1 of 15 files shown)


esoteric-ephemera:
@tsmathis and @tschaume  ok to merge? There's something up with the latest pymatgen release (2025.1.4) that is also causing problems in atomate2 - need to investigate that to release the version pin

emmet-api/emmet/api/routes/materials/materials/query_operators.py
```diff
@@ -409,12 +409,16 @@ def query(
             comp_red = comp.reduced_composition.items()
 
             for i, j in comp_red:
+                # The keys of pymatgen's Composition can be Element, Species, or DummySpecies
+                # Element and Species both have a name attr, DummySpecies doesn't by default
+                # This bothers mypy a lot, so we placate it here - all three have __str__ methods:
+                spec_name = str(getattr(i, "name", None) or i)
```
tschaume, line 415: The fallback to `i` here is for `DummySpecies`?

tschaume (approved):

emmet-api/emmet/api/routes/materials/materials/query_operators.py
esoteric-ephemera: Yeah exactly - it's completely unnecessary for MP's data (we only use `Element` and `Species`), but accommodates any of the possible composition key types

PR #1190, opened 2025-02-14
[WIP] Add emmet-archival namespace package

esoteric-ephemera:
## Summary

Adds an `archival` namespace package to `emmet` to establish MP's data archival formats moving forward. This package should be flexible enough to permit new data formats as they become relevant (e.g., data from phonopy, LOBSTER, etc.)

- Base `Archiver` class that defines a schema for how data is to be archived / retrieved from HDF5 + zarr files
- VASP `RawArchive` class:
  - Stores only relevant VASP files as string / bytes archives in hierarchical format
  - Automatically strips copyright protected info from VASP files
  - Constructs `TaskDoc` from HDF5/zarr data
- VASP `VolumetricArchive` + separate volumetric data classes: formats for converting `CHGCAR`-structured files, bandstructures, and DOS to HDF5/zarr
- `TrajArchive` for storing trajectory information. Agnostic to traj format type

## To Do's:
- Tests tests tests
- Extend `TrajArchive` to strip ionic step info from a `TaskDoc`
- Test formats on a subset of VASP data on S3 to estimate storage size, partial retrieval, etc.
- Parquet extensions?


emmet-archival/emmet/archival/vasp/raw.py
```diff
@@ -0,0 +1,173 @@
...
+            raise SystemError(f"Directory {calc_dir} does not exist!")
+
+        metadata: dict[str, str | dict[str, str]] = {
+            "calc_dir": str(calc_dir),
+            "file_paths": {},
+            "md5": {},
+        }
+
+        parsed_objects: dict[str, Any] = {}
+        for calc_type, file_list in VASP_RAW_DATA_ORG.items():
```
kbuma, line 55: As I work on the new CLI, I'm trying to understand how best to gather raw data for a submission. I've found that the files used for building tasks  from raw data (and for the validator) are being discovered using [def _find_vasp_files(](https://github.com/materialsproject/emmet/blob/main/emmet-core/emmet/core/tasks.py#L1055)
This looks to be a departure from that. Is this the way we want to proceed going forward?

Also, whichever way we choose, I'd like to expose the method for finding the files so it can be reused in the new CLI.

emmet-archival/emmet/archival/trajectory.py
```diff
@@ -0,0 +1,493 @@
...
+    energy = "energy"
+    forces = "forces"
+    stress = "stress"
+    stress_voigt = "stress_voigt"
+    magmom = "magmom"
+    charge = "charge"
+
+
+@dataclass
+class TrajArchive(Archiver):
```
kbuma, line 46: Consider calling this TrajectoryArchive instead for naming consistency - every other Archive class uses the full terms.

emmet-archival/emmet/archival/base.py
```diff
@@ -0,0 +1,318 @@
...
+        fmt = infer_archive_format(archive_path)
+        if fmt in (ArchivalFormat.HDF5, ArchivalFormat.ZARR):
+            with Archiver.load_archive(archive_path, fmt=fmt, mode="r") as _f:
+                return cls._extract_from_hdf5_like(_f, *args, **kwargs)
+        elif fmt == ArchivalFormat.PARQ:
+            return cls._extract_from_parquet(archive_path, *args, **kwargs)
+
+    @contextmanager
+    @staticmethod
+    def load_archive(
```
kbuma, line 144: This method is doing different things based on the mode parameter. It's confusing to call it load_archive because of that.
Suggest making the current implementation a private class method that's for getting a handle to a HPDF5/ZARR file and creating public class methods for the 2/3 use cases along the lines of get_new_archive_for_writing, get_archive_for_append, get_archive_for_reading

esoteric-ephemera: Good call, will rename to `_open_hdf5_like`. Do you think it's worth exposing this method to the user as opposed to leaving as an underscored method? 

This exists mostly to make the file I/O APIs the same between HDF5 and zarr

kbuma: I'd say start with it underscored - I think it'll be clearer for the user to use the helper methods that set the mode for them.

emmet-archival/emmet/archival/base.py
```diff
@@ -0,0 +1,318 @@
...
+        archive_name: str | Path,
```
kbuma, line 145: Can you please make this also accept zarr.storage.StoreLike ; I believe that will then let us use fsspec to store/read directly to S3

emmet-archival/emmet/archival/base.py
```diff
@@ -0,0 +1,318 @@
...
+        """Extract data from an HDF5-like file."""
+        raise NotImplementedError
+
+    @classmethod
+    def _extract_from_parquet(cls, archive_path: str | Path, *args, **kwargs) -> Any:
+        """Extract data from a parquet file."""
+        raise NotImplementedError
+
+    @classmethod
+    def extract(cls, archive_path: str | Path, *args, **kwargs) -> Any:
```
kbuma, line 132: Can you please make archive_path also accept zarr.storage.StoreLike ; I believe that will then let us use fsspec to store/read directly to S3

emmet-archival/emmet/archival/base.py
```diff
@@ -0,0 +1,318 @@
...
+        """Append data to an existing HDF5-like file group."""
+        raise NotImplementedError
+
+    def _to_parquet(self, file_name: PathLike, **kwargs) -> None:
+        """Write data to a parquet file."""
+        raise NotImplementedError
+
+    def to_archive(
+        self,
+        file_name: PathLike = "archive.h5",
```
kbuma, line 89: Can you please make this also accept zarr.storage.StoreLike ; I believe that will then let us use fsspec to store/read directly to S3

emmet-archival/emmet/archival/base.py
```diff
@@ -0,0 +1,318 @@
...
+                _dtype = pd.Float64Dtype()
+            data[k] = pd.array(v, dtype=_dtype)
+        data["_attributes"] = {
+            "lattice": structure.lattice.matrix,
+            "charge": structure.charge,
+        }
+        return data
+
+    @staticmethod
+    def columnar_to_structure(column_dict: dict[str, Any]) -> Structure:
```
kbuma, line 258: Suggest renaming method or changing type. Since you have a structure_to_columnar method that returns a DataFrame best practice would be that the columnar_to_structure method takes a DataFrame as the parameter.

emmet-archival/emmet/archival/vasp/raw.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+
+    @classmethod
+    def from_directory(cls, calc_dir: str | Path, **kwargs) -> Self:
+        calc_dir = Path(calc_dir).resolve()
+        if not calc_dir.exists():
+            raise SystemError(f"Directory {calc_dir} does not exist!")
+
+        file_paths: dict[str, FileMetadata] = {}
+        for file_meta in discover_vasp_files(calc_dir):
```
kbuma, line 53: Could you please extract a method here that takes a list of FileMetada and returns the file_paths dictionary such that the keys are constructed in the way that RawArchive expects. This will allow the CLI to utilize RawArchive rather easily.

esoteric-ephemera: Moved to `emmet.archival.vasp.raw.raw_archive_hierarchy_from_files`

emmet-archival/emmet/archival/base.py
esoteric-ephemera: Good idea - tweaked this to be consistent

esoteric-ephemera:
@kbuma - need to check that the zarr store integration is correct but reading through their docs it *seems* ok. Will check with some small examples

kbuma:
@esoteric-ephemera I think I sent you down the wrong path last week with the zarr store integration with fsspec

the following style of call works (if s3fs is installed) so the extra parameter is not necessary and we can use the archive parameter as you had it before - only thing I would encourage is to use zarr.open_group rather than zarr.open since we are always using a group in our use cases

zarr.open_group("s3://bucket/mygroup.zarr", mode="w")

esoteric-ephemera:
@kbuma : integrated your `CalculationLocator` with `emmet.core.vasp.utils`, and reworked the vasp validation temporarily just to ensure that the document fields are correctly populated

Since the validation is tested externally, I don't think that we need to have as in-depth tests in emmet

esoteric-ephemera:
@tschaume to keep other PRs moving (like Karlo's new CLI) and to ensure we're making more easily-reviewed changes to new namespaces, going to checkpoint the archival package here. More updates will go into a separate PR

Most of the changes to emmet-core are additive. The only thing you might want to check @tsm is that the phonon bandstructure and DOS schemas still are OK with OpenData. These just inherit from a new base class which schematizes generic bandstructure and DOS objects

tschaume (changes requested):
Wow, great work! I only have nit-picky comments to go through.

emmet-api/setup.py
```diff
@@ -23,6 +23,7 @@
         "asgi-logger",
         "pymatgen-analysis-alloys>=0.0.3",
         "pymatgen-analysis-defects>=2024.10.22",
+        "pymatgen-io-validation>=0.1.0",
```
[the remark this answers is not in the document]
esoteric-ephemera: It should be - will double check that upgrade dependencies picks that up. It didn't for `emmet-archival` unless I explicitly set this in `setup.py` and `pyproject.toml`, but that might be the two conflicting

emmet-archival/emmet/archival/vasp/raw.py
```diff
@@ -0,0 +1,380 @@
...
+                arch = Path(file_arch)
+                base_group = str(arch.parent)
+                with zopen(file_meta.path, "rb") as vhf_b, h5py.File(
+                    vhf_b, "r"
+                ) as vh5f:
+                    if base_group not in group:
+                        group.create_group(base_group)
+                    vh5f.copy("/", group[base_group], name=arch.name)
+                if "vaspout" in file_arch:
+                    ppath = str(arch / "input/potcar/content")
```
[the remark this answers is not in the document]
esoteric-ephemera: This is the syntax for hierarchies in HDF5 and zarr, should be the same for unix / windows

emmet-archival/emmet/archival/vasp/volumetric.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def to_arrow(self) -> pa.Table:
+        """Convert DOS archive to arrow table."""
+        return self.dos.to_arrow()
+
+    @classmethod
+    def from_arrow(cls, table: pa.Table) -> Dos | CompleteDos:
+        """Extract a pymatgen DOS from an arrow table."""
+        return ElectronicDos.from_arrow(table).to_pmg()
+
```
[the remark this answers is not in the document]
esoteric-ephemera: All the `Archive`-derived objects have an `extract` method which internally calls `from_arrow` to read parquet files. So for right now, the signature of `from_arrow` for an `Archive` returns the pymatgen object, whereas the schema for the pymatgen object (`ElectronicDos` here) returns the pydantic schema from `from_arrow`.

Can change this so we always return the pydantic schema rather than pymatgen. I think that's the direction we're moving in right?

emmet-archival/pyproject.toml
```diff
@@ -0,0 +1,106 @@
...
+
+[tool.coverage.report]
+show_missing = true
+skip_covered = true
+
+[tool.pydocstyle]
+convention = 'google'
+match = '^pymatgen/(?!_).*\.py'
+inherit = false
+add-ignore = "D107, "
```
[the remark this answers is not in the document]
esoteric-ephemera: Right! I didn't copy the structure of this....

emmet-core/emmet/core/vasp/validation.py
```diff
@@ -48,435 +55,124 @@ class DeprecationMessage(DocEnum):
     UNKNOWN = "U001", "Cannot validate due to unknown calc type"
 
 
-class ValidationDoc(EmmetBaseModel):
+class ValidationDoc(VaspValidator, EmmetBaseModel):
```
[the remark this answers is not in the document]
esoteric-ephemera: Part of the archival PR, Karlo needs them for the new CLI. The builders still use the current validation method in `validation_legacy`

emmet-core/emmet/core/band_theory.py
```diff
@@ -0,0 +1,355 @@
...
+
+    @model_serializer
+    def serialize_pmg(self) -> dict[str, Any]:
+        """Serialize pymatgen objects to dicts."""
+        config = {k: getattr(self, k, None) for k in type(self).model_fields}
+        if config.get("structure"):
+            config["structure"] = config["structure"].as_dict()  # type: ignore[union-attr]
+        return config
+
+    @requires_arrow
```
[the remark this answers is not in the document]
esoteric-ephemera: Still optional in emmet-core but required in emmet-archival for simplicity

emmet-core/emmet/core/vasp/calc_types/utils.py
```diff
@@ -68,7 +73,7 @@ def task_type(
     incar = inputs.get("incar", {})
     kpts = inputs.get("kpoints") or {}  # kpoints can be None, then want a dict
 
-    if not isinstance(kpts, dict):
+    if isinstance(kpts, Kpoints):
```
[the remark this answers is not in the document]
esoteric-ephemera: I think I switched this because of mypy - basically the check should just be that `kpts` has `as_dict` as an attr and then runs if needed

PR #1194, opened 2025-02-18
CI fix

esoteric-ephemera:
Very small fix to patch CI segfaults - seems like the issue is with newer versions of the `openff` libraries. Pinning them to the versions we use in `atomate2`'s CI

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Install all conda requirements
         shell: bash -l {0}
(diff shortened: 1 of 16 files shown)


esoteric-ephemera:
@tschaume: had to put a pin on the `openff` dependencies as well as `pymatgen`. For now, I think we need `pymatgen<=2025.1.9` to ensure that the new bandgap parsing doesn't affect tests. Would like to address that change in a separate PR

tschaume:
Thanks for tracking this down! Can we put the pymatgen pin also into the setup.py for `emmet-core`? We'd want to avoid pymatgen being bumped when I run the action to upgrade dependencies next time.

tschaume:
Yes, looks good. Thanks!

PR #1206, opened 2025-03-12
New trajectory class with parquet rountrip functionality

esoteric-ephemera:
In support of removing ionic step / calcs reversed info from MP's mongo task collection, this adds a `Trajectory` class which interfaces with pyarrow/parquet, pymatgen's `Trajectory`, and ASE's `Trajectory`. Verified that parquet rountrip works perfectly (model-dumped hashes of emmet `Trajectory` objects are identical before and after parquet conversion).

Since the site still needs energy convergence info, using parquet lets us partially retrieve the energy data from the trajectory

This is a middle-ground solution until the [emmet-archival](https://github.com/materialsproject/emmet/pull/1190) PR is ready


esoteric-ephemera:
@tschaume @yang-ruoxi : should be ready to test with the trajectories endpoint, roundtrip is working fine. Only additions I might want to make are supporting site properties, but I don't think we have any in the ionic steps (e.g., selective dynamics, magmoms, and velocities tags)

tsm:
Dear people of project tschaume (especially Aaron),

please do not call me (my github account is tsm), I am not member of your project.

Best regards,
Tsm

tsm:
project emmet*

tsmathis:
Sorry about that, looks like my github username was mixed up with my slack username. Unsubscribing from the issue should prevent any further notifications from reaching you

tsmathis:
re: arrow + parquet writing, is the long term intention to keep writing individual trajectory objects to individual parquet files?

esoteric-ephemera:
No this is a short-term solution: To build a performant index for the task collection, removing the ionic steps / calcs reversed helps a lot (reduces the task collection size by half). We still want to serve that info up, and need the total energy by ionic step for the [convergence graph](https://next-gen.materialsproject.org/materials/mp-149/tasks/mp-1947498) in the website task view

Serving up individual parquet files permits partial retrieval of energy data by task_id, and also lets users retrieve full trajectory info

emmet-core/emmet/core/trajectory.py
```diff
@@ -0,0 +1,424 @@
...
+            if hasattr(v, "__len__")
+        }
+        pa_config["elements"] = pa.array(
+            [self.elements for _ in range(self.num_ionic_steps)]
+        )
+
+        pa_table = pa.table(pa_config)
+        if file_name:
+            with zopen(str(file_name), "wb") as f:
+                pa_pq.write_table(pa_table, f)
```
tsmathis, line 293: Correct me if I'm wrong, but the compression formats that are parquet compatible (‘SNAPPY’, ‘GZIP’, ‘BROTLI’, ‘LZ4’, ‘ZSTD’) don't mesh with `zopen`'s formats.

 I see in the test files there is a `.gz` extension for the test parquet file, I'm guessing this is the only format that would work with `zopen`?
 
 I would opt towards dropping monty here and sticking with pyarrow.parquet's read/write behavior and support all the compression types that are parquet compatible and have the default be the default compression format for write_table, i.e., `snappy`: [pyarrow.parquet.write_table](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html#pyarrow-parquet-write-table).

esoteric-ephemera: Yeah `monty` doesn't support these - switched to default pyarrow compression, thanks for the suggestion!

tschaume:
Thank you @esoteric-ephemera! Once we get around to trying to put the new Task API into production, I'll have to double-check whether it's more performant to put [the data needed](https://github.com/materialsproject/web/blob/8c1be2703e1e265ef87e5754e1058e012221e155/mp_web/pages/apps/materials/task_details.py#L296-L327) for the convergence graph back into the collection or retrieve them through parquet files (preferably the same @tsmathis is using for the builders).

esoteric-ephemera:
@tschaume, when the `requirements_*.txt` files are auto-generated, it looks like the package itself [isn't being excluded](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/emmet-core/requirements/ubuntu-latest_py3.10_extras.txt#L81). Causes problems with CI if version pins change / probably want to ensure that only the current version is tested in CI

tschaume:
@esoteric-ephemera This comes down to [this line](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/.github/workflows/testing.yml#L94) in the workflow file. I can't remember why I decided to only install an editable version of emmet-core for testing the other emmet namespace packages. We could change that line to `matrix.package == "emmet-core"` or maybe better remove the line entirely so that always the latest version is tested.

tschaume:
PS: actually I remember :) [The next step](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/.github/workflows/testing.yml#L97-L99) in the action should install the current emmet-core over any version defined in the requirements file.

esoteric-ephemera:
OK makes sense, but for releasing upper version pins on dependencies, this can cause CI issues

tschaume:
If the upper version pin is on dependencies other than `emmet-*`, you'd update the setup.py in your branch, run the `upgrade dependencies` action from your branch which will create an `upgrade-dependencies` branch (and PR) that contains the new set of dependencies, and then merge that into your branch. Hope that makes sense.

PR #1207, opened 2025-03-14
Fix LDA pseudo parsing and related parsing bugs

### LINKED ISSUE #847: Bug when parsing LDA pseudopotential
### Problem

Trying to read in a VASP run with an LDA (i.e. `PAW` but not `PAW_PBE`) pseudopotential causes a POTCAR parsing issue on emmet-core 0.69.6 and pmg 2023.10.4.

```python
from emmet.core.tasks import TaskDoc

TaskDoc.from_directory('.')
```

```
File /global/common/software/matgen/rosen/miniconda/envs/quacc/lib/python3.10/sn, additional_fields, volume_change_warning_tol, **vasp_calculation_kwargs)
    488 if vasp_objects:
    489     included_objects = list(vasp_objects.keys())
    491 doc = cls.from_structure(
    492     structure=calcs_reversed[0].output.structure,
    493     meta_structure=calcs_reversed[0].output.structure,
    494     include_structure=True,
    495     dir_name=dir_name,
    496     calcs_reversed=calcs_reversed,
    497     analysis=analysis,
    498     transformations=transformations,
    499     custodian=custodian,
    500     orig_inputs=orig_inputs,
    501     additional_json=additional_json,
    502     icsd_id=icsd_id,
    503     tags=tags,
    504     author=author,
    505     completed_at=calcs_reversed[0].completed_at,
--> 506     input=InputDoc.from_vasp_calc_doc(calcs_reversed[-1]),
    507     output=OutputDoc.from_vasp_calc_doc(
    508         calcs_reversed[0],
    509         vasp_objects.get(VaspObject.TRAJECTORY),  # type: ignore
    510     ),
    511     state=_get_state(calcs_reversed, analysis),
    512     run_stats=_get_run_stats(calcs_reversed),
    513     vasp_objects=vasp_objects,
    514     included_objects=included_objects,
    515     task_type=calcs_reversed[0].task_type,
    516 )
    517 doc = doc.model_copy(update=additional_fields)
    518 return doc

File /global/common/software/matgen/rosen/miniconda/envs/quacc/lib/python3.10/s
    195 if xc:
    196     xc = xc.upper()
--> 198 pot_type, func = calc_doc.input.potcar_type[0].split("_")
    199 func = "lda" if len(pot_type) == 1 else "_".join(func)
    200 pps = Potcar(pot_type=pot_type, functional=func, symbols=calc_doc.input

ValueError: not enough values to unpack (expected 2, got 1)
```

I can't share a minimal example here due to proprietary reasons, but I can send one over Slack.

### Proposed Solution

If the string label can't be split, assume LDA? I think?

### Alternatives

_No response_

esoteric-ephemera:
Closes #847: parsing LDA pseudopotential calcs failed using `TaskDoc`. Also fixed another bug that would render `TaskDoc.input.pseudo_potentials` as:
```
Potcar(pot_type='PAW', functional='P_B_E', symbols=['PAW_PBE'])
```
rather than
```
Potcar(pot_type='PAW', functional='PBE', symbols=['Si'])
```
(double-checked: the tasks on MP all have the former `Potcar` style if that field is populated).

Both of the previous parsing issues stem from the fact that we don't parse the majority of POTCAR info from the POTCAR, but the vasprun.xml. Parsing directly from POTCAR would let us access these attributes from the individual `PotcarSingle` objects (`potential_type`, `LEXCH`, and `symbol`, respectively)

I can change the parsing to prefer POTCAR-derived attributes if that's desirable moving forward. Also forward-looking in case we change to parsing from vaspout.h5, which contains the full POTCAR as a str


tschaume:
Thanks @esoteric-ephemera ! PR looks good and is ok to merge I think.

I think I'm in favor of parsing directly from the POTCAR in the future. Couple of questions to clarify:

- Are there any scenarios in which the info from the POTCAR could be different from what's saved in the vasprun at the end of a calculation?
- Would we be able to parse the `PotcarSingle` from the original POTCAR as well as the non-proprietary POTCAR hash we'll be saving on OpenData?
- We would probably also have to remove the full POTCAR string and replace with the hash in `vaspout.h5`, right? Is there any proprietary POTCAR info stored in the current `vasprun.xml`?

esoteric-ephemera:
> Are there any scenarios in which the info from the POTCAR could be different from what's saved in the vasprun at the end of a calculation?

The info should be the same as in vasprun.xml, but there's less info stored in vasprun.xml overall. I think it's just the "TITEL" fields stored in vasprun.xml and the element symbols

> Would we be able to parse the PotcarSingle from the original POTCAR as well as the non-proprietary POTCAR hash we'll be saving on OpenData?

Some of the POTCAR hashes are the same because they weren't updated in between releases (like between PBE_52 and PBE_54 releases) so sort of - we can give a set of guesses for what the POTCAR was.

In the [new validation scheme for POTCARs](https://github.com/materialsproject/pymatgen-io-validation/blob/c14f403772a0558d0b911b37e5bbb36a080de4bb/pymatgen/io/validation/check_potcar.py#L61), we build a subset of POTCAR hashes that are acceptable, and use those to validate a guess POTCAR or its hash. That gets around the issue of needing to ID the POTCAR uniquely

> We would probably also have to remove the full POTCAR string and replace with the hash in vaspout.h5, right? Is there any proprietary POTCAR info stored in the current vasprun.xml

We do but there's a [feature](https://github.com/materialsproject/pymatgen/blob/eec44223101d24c690e001ff3cf8d3dfa00e7e02/src/pymatgen/io/vasp/outputs.py#L6407) in the vaspout.h5 parser I added to pymatgen that does just that. vasprun.xml has no copyrighted info

tschaume:
ok, perfect, thanks @esoteric-ephemera! I'll merge this PR and release an `rc` version for @Andrew-S-Rosen to use/test. We'll probably start separate PRs for preferring POTCAR-derived attributes and for potential h5 parsing when it's a good time to tackle these. The former might come into play as soon as we start parsing the recomputed R2Scan data.

PR #1209, opened 2025-03-26
Project `run_type` to ensure `entries` field is correctly constructed

mkhorton:
There seems to be a bug in `MaterialsBuilder` where `run_type` is not projected, leading to a `"run_type": None` in the corresponding `entry`. Following this through the builder pipeline will result in those entries not being seen as valid by the compatibility scheme.

It's unclear to me when this bug was introduced, I assume it hasn't hit any Materials Project production builds because it would have been very noticeable.

Thanks to @basveeling for identifying and fixing this bug.

### DIFF
--- a/emmet-builders/emmet/builders/vasp/materials.py
+++ b/emmet-builders/emmet/builders/vasp/materials.py
@@ -186,6 +186,7 @@ class MaterialsBuilder(Builder):
             # needed for run_type and task_type
             "calcs_reversed.input.parameters",
             "calcs_reversed.input.incar",
+            "calcs_reversed.run_type",
             "orig_inputs",
             "input.structure",
             # needed for entry from task_doc


tsmathis:
likely related to the `TaskDocument` (atomate) -> `TaskDoc` (atomate2) move

`run_type` was a `@property` in `TaskDocument` so it would get populated at run time, likely here: [.from_tasks()](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/emmet-core/emmet/core/vasp/material.py#L80)

`run_type` is an optional pydantic field in `TaskDoc`'s calcs_reversed field: [TaskDoc.calcs_reversed](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/emmet-core/emmet/core/tasks.py#L346)->[Calculation.run_type](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/emmet-core/emmet/core/vasp/calculation.py#L652)

I do however remember a [post init](https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/emmet-core/emmet/core/tasks.py#L454) being put on `TaskDoc` to correctly populate the run/calc/task_type attributes during deserialization #1033, I think the following projected fields should cause that post init to run: 
https://github.com/materialsproject/emmet/blob/ab1e34a74ec75e958ee3f0fe59333fef7e52186c/emmet-builders/emmet/builders/vasp/materials.py#L186-L190

Just a guess here, @mkhorton, what is the schema for your tasks collection? `TaskDocument` or `TaskDoc`?

mkhorton:
Thanks for the added context Tyler. To be honest, I am actually not sure off-hand, the original calculation could be either. Any older calculations I was looking at were migrated to the newer `TaskDoc` to ensure consistency but perhaps some bug was introduced during in the migration.

Will leave it to your best judgement for whether to merge, I'm not sure if this is something that would affect other users or not.

tsmathis:
No problems here with the added projection if it solves your immediate issue, will merge. I think we're due for a release too.

Was mostly thinking "aloud" because I thought the `TaskDoc` `run_type`, `calc_type`, and `task_type` issues had been addressed and this may have been hinting at a different underlying issue.

I do also remember MP had to re-migrate our `tasks` collection again after #1033 at @esoteric-ephemera's recommendation to ensure our atomate2 TaskDoc's had the correct run/calc/task_types. It may be worth spot-checking your `tasks` collection to see if those fields are as you expect. Happy to share my migration script if it would be helpful.

tsmathis:
@mkhorton v0.84.6 is live with your patch

PR #926, opened 2024-01-18
[WIP] Large structural changes to MPcules

espottesmith:
Meant to do this piecemeal, but there kept being things that I wanted/needed.  Also, the main point was to make changes/fixes to the API, and that's actually one of the things I still need to mess with/test. Anyways, here we are.

## Contributor Checklist

- [X] I have broken down my PR scope into the following TODO tasks
   -  [X] Add new document containing molecular dipole/multipole data and associated builders
   -  [X] On @munrojm's suggestion, changed the has_props field in the molecule summary doc to aid in querying
   -  [X] Remove probably unnecessary query operators
   -  [X] Switch molecule element querying to be based on composition, which should be more efficient?
   -  [X] Add dipoles/multipoles to the information returned in a trajectory query. Also improved testing for the trajectory query operator
   -  [X] Add three-center (3C) bonds and 3-center 4-electron hyperbonds to the orbital doc/builder. Thanks @samblau for making the changes to pymatgen necessary to parse this data
   -  [X] Change all molecule property builders to break down documents based on hash, rather than formula. Even for the small and quick property builders, the current scale of the data (~4M tasks, ~600k molecules) made separation by formula problematic.
   -  [x] Reformat some data to remove lists (where possible) and make querying on summary document faster
   -  [ ] Make some changes to NBO-based bonding, based on some issues identified by @samblau and @orionarcher
   -  [ ] (Potentially) add string reps (InChI and SMILES) to all molecules docs, rather than just tasks (for SMILES) and molecules/summary (for InChI)
- [X] I have run the tests locally and they passed. (I never run ALL the tests, cause that takes a hot minute, but I did run all of the tests that I messed with!)
- [X] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR

PR #1210, opened 2025-03-28
Fix SyntaxWarning: invalid escape sequence '\D'

ktns:
Hi. I'm using python3.12 and emmet-core and getting SyntaxWarnings which is a little bit annoying.

I think it can be avoided by simply putting `r` before the quotes.


tsmathis:
Hi @ktns, thanks for contributing! Good catch.

I'm guessing #926 would have gotten rid of this, but that was reverted unfortunately. Hopefully that effort will be revived soon.

tsmathis:
@ktns, your fix is released as part of v0.84.6

PR #1218, opened 2025-04-01
Automated dependency upgrades

tschaume:
Full log: https://github.com/materialsproject/emmet/actions/runs/14207563906


esoteric-ephemera:
@tschaume you might need to rerun the action, I was trying to see if the newer `robocrys` removed the circular dependency tree and didn't realize running the action on my branch would merge in commits from that branch

The circular dependencies are gone so that's good, will merge that fix separately

tsmathis:
I've been triggering the upgrade workflow manually, so after your fix is in I'll re-run everything and see where things stand

PR #1220, opened 2025-04-02
Tweak potcar validation

esoteric-ephemera:
Changes how POTCARs are validated slightly. It seems that **most** if not all of the PBE_54 POTCARs are identical to the PBE_54_W_HASH variants. The latter set includes extra tags: COPYR (copyright) and SHA256 (hash of the file). The actual data contents appear to be the same

The POTCAR validation now defaults to excluding these COPYR and SHA256 fields to perform validation

Also bumps robocrys to 0.2.11 to remove circular software dependence

Noting this here for future reference: There appear to be extra GW-related tags in at least these PBE_54_W_HASH POTCAR: `['C','Na_pv','Mg_pv','Ca_sv','Cl',O','Li_sv']`. Can't say with certainty if those differences are important for regular DFT calcs, but they won't validate against PBE_54 POTCARs for the time being

### DIFF
--- a/emmet-core/emmet/core/openmm/tasks.py
+++ b/emmet-core/emmet/core/openmm/tasks.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import io
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional, Union, TYPE_CHECKING
 
 import openmm
 import pandas as pd  # type: ignore[import-untyped]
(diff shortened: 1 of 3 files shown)


tsmathis:
@esoteric-ephemera, can you take care of the mypy errors when you have a chance?

esoteric-ephemera:
@tsmathis should be good to go once the tests pass, will merge in

PR #1226, opened 2025-04-26
Rearrange TaskDoc top level fields, fix from_directory parsing

esoteric-ephemera:
* Ensure that `TaskDoc.from_directory` populates the `potcar_spec` of `orig_inputs` and not `potcar` field
* Make `orig_inputs` and `input` both `CalculationInput` objects, rather than derived classes
* Mark derived `OrigInputs` and `InputDoc` classes as deprecated, and move missing top-levels to properties on `CalculationInput`
* Make testing utils importable `emmet.core.testing_utils`
* Add structure similarity calculation (prep for new builder) + test
* Ensure all of emmet-core passes mypy (removed blanket file ignores + cleaned these up)

Had to rearrange a few builder logic pieces to accommodate this change. Tag @tsmathis and @tschaume  for review
(Apologies to the unrelated person I mistakenly tagged again.)

Tl ; dr : streamline various inputs field of TaskDoc, make them a little more consistent


tsm:
@esoteric-ephemera Please stop calling me by `@`. I am not collaborator of this project!

tsmathis:
LGTM

Making sure I have this right on the database side for schema consistency: 

1. I can(should) drop the following fields: [`orig_inputs.poscar`, `input. pseudo_potentials`, `input. xc_override`, `input. is_lasph`, `input. magnetic_moments`]
2. All docs parsed with `TaskDoc.from_directory(..)` prior to PR  incorrectly have a list of `potcar_spec`s in the `orig_inputs.potcar` field, those should be moved to `orig_inputs.potcar_spec`
3. All docs that have an emmet `Potcar` model dict in `orig_inputs.potcar` should take the list of strings in `orig_inputs.potcar.symbols` and set that as `orig_inputs.potcar` field

We should be prepared to offer migration guidance if users start bringing up issues with their own tasks collections.

Since our tasks collection is a mix of migrated `TaskDocuments` and newly parsed `TaskDocs`, I've found these three cases in the `orig_inputs.potcar` field:
1. emmet `Potcar` model dicts (most of the collection)
2. lists of potcar symbol strings (<1%)
3. lists of `potcar_specs` (all new entries, <10%)

@tschaume, I'll have to make sure we make this migration before the next deployment that includes an `emmet` upgrade. These model changes will break deserialization for the tasks endpoint with the current version of the tasks collection.

esoteric-ephemera:
@tsmathis a couple changes:
> 1. I can(should) drop the following fields: [orig_inputs.poscar, input. pseudo_potentials, input. xc_override, input. is_lasph, input. magnetic_moments]

We should be able to drop these. Some data might be lost if these fields alone are populated rather than the preferred attrs. For example, if orig_inputs.poscar is populated, but orig_inputs.structure is not, you'd also need to do the following:
```
orig_inputs.structure = orig_inputs.structure or orig_inputs.poscar.structure
if inputs.is_lasph and not inputs.parameters.get("LASPH"):
    inputs.parameters["LASPH"] = inputs.is_lasph
```

The other two look good to me!

Before we migrate, we should probably iterate on this a bit / maybe get community input

tsmathis:
Makes sense, though the best path forward imo on the emmet side is ensuring there is one path for setting all these values at parsing time. And then any inconsistencies in the database with prior entries we just need to run migrations to conform to the updated schema.

Given your changes in this PR, are there any times this would be the case when calling `TaskDoc.from_directory(..)` to create a new TaskDoc?

>attrs. For example, if orig_inputs.poscar is populated, but orig_inputs.structure is not, you'd also need to do the following:
> 
> ```
> orig_inputs.structure = orig_inputs.structure or orig_inputs.poscar.structure
> if inputs.is_lasph and not inputs.parameters.get("LASPH"):
>     inputs.parameters["LASPH"] = inputs.is_lasph
> ```

Can `input.structure` ever not be set for a new `TaskDoc`?
Is it even possible to set `is_lasph` on `inputs` without `"LASPH"` in the input parameters?

esoteric-ephemera:
`TaskDoc.from_directory` should parse correctly, those other suggestions were just for database / user migration.

>  Is it even possible to set is_lasph on inputs without "LASPH" in the input parameters?

It would be strange for sure, but maybe in the really old TaskDocument-derived tasks?

tsmathis:
The last thing I would consider is the potcar checks in task validation: https://github.com/materialsproject/emmet/blob/2aa5527da1b041f5b4ba430785c5fbb3aecad145/emmet-core/emmet/core/vasp/validation.py#L358

Will this change start breaking things there?

And on that note, all of the TaskDocs migrated from old TaskDocuments of course do not have the potcar_specs field populated, is it possible to backfill the potcar_specs for those older entries? Or is that a `.from_directory()` operation only?

emmet-core/emmet/core/tasks.py
```diff
@@ -1004,13 +895,10 @@ def _parse_orig_inputs(
...
             if f"{name}{suffix}" in str(filename):
-                if name == "POTCAR":
-                    # can't serialize POTCAR
-                    orig_inputs[name.lower()] = PotcarSpec.from_potcar(
-                        vasp_input.from_file(filename)  # type: ignore[attr-defined]
-                    )
-                else:
-                    orig_inputs[name.lower()] = vasp_input.from_file(filename)  # type: ignore[attr-defined]
+                file_suffix = "_spec" if name == "POTCAR" else ""
+                orig_inputs[f"{name.lower()}{file_suffix}"] = vasp_input.from_file(
```
tsmathis, line 899: this panics on the "POTCAR" iteration -> `AttributeError: from_file`
`PotcarSpec` has no `from_file` method, right?

Keeping the original if: ... else: ...
but changing this works:
```python
                    orig_inputs["potcar_spec"] = PotcarSpec.from_potcar(  # <- instead of name.lower()
                        vasp_input.from_file(filename)  # type: ignore[attr-defined]
                    )
```

PotcarSpec has a `.from_potcar()` method that returns a list of `PotcarSpec`s, but I think that would be same as the list comp above?

tsmathis: ah great, sorry I missed that 👍

PR #1232, opened 2025-05-16
Slim version of TaskDoc for production

tsmathis:
~~Will be considered experimental until performance + compatibility with builder/api client can be tested~~
~~@esoteric-ephemera, check if the field split I made here makes sense. Can worry about model validators and a parsing method later after I've tested the basic schema works for the build pipelines + api~~
~~Based this branch off your `alpha_id` branch for the updated `task_id` typing.~~
~~Also makes more sense for the top level `output` field to be the calc output right?~~

Now ready, broad strokes:
- `CoreTaskDoc` represents a single calculation, can't parse multiple calcs
    - `TaskDoc` parsing behavior remains unchanged, can parse multiple calcs in a single dir
- `TaskDoc` is now a superset of `CoreTaskDoc`
    - `CoreTaskDoc` is strict on typing
    - no `calcs_reversed` in `CoreTaskDoc` or other fields that shouldn't be stored at the database level
    - `CoreTaskDoc.output` is an instance of `emmet.core.tasks.calculation.CoreCalculationOutput` for parity with `(Core)TaskDoc.input` being an instance of `emmet.core.tasks.calculation.CalculationInput` -> i.e., all relevant quantities  at surface of document, not nested
- `CalculationOutput` is now a superset of `CoreCalculationOutput`
- `TaskDoc.output` (`emmet.core.tasks.OutputDoc`) shadows `CoreTaskDoc.output` (`emmet.core.vasp.calculation.CoreCalculationOutput`)

Notes:
- `TaskDoc` behavior and parsing shouldn't be effected at all (no downstream breaks)
- Parsing `CoreTaskDoc` from a directory yields a tuple of (CoreTaskDoc, CoreTrajectory)
   - on the caller to handle storing their trajectory and task document (and making sure `CoreTaskDoc.task_id` and `CoreTrajectory.identifier` are set correctly)

- [x] Tests are gtg, the slim models (`CoreTaskDoc` & `CoreCalculationOutput`) are just slices of the existing models

### DIFF
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -383,7 +383,7 @@ class CoreTaskDoc(StructureMetadata):
             vasp_objects=vasp_objects,
         )
 
-        props = {
+        props: dict[str, Any] = {
             "structure": [],
             "cart_coords": [],
             "electronic_steps": [],
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -313,7 +313,7 @@ class CoreTaskDoc(StructureMetadata):
                 for char in batch_id
                 if (not char.isalnum()) and (char not in {"-", "_"})
             )
-            if len(invalid_chars) > 0:
+            if invalid_chars:
                 raise ValueError(
                     f"Invalid characters in batch_id: {' '.join(invalid_chars)}"
                 )
--- a/emmet-core/emmet/core/tasks.py
+++ b/emmet-core/emmet/core/tasks.py
@@ -4,16 +4,16 @@ from __future__ import annotations
 
 import logging
 import re
+import warnings
 from collections.abc import Mapping
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
-import warnings
 
 import numpy as np
 from monty.json import MontyDecoder
 from monty.serialization import loadfn
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
 from pymatgen.analysis.structure_analyzer import oxide_type
 from pymatgen.core.structure import Structure
 from pymatgen.core.trajectory import Trajectory
@@ -21,8 +21,9 @@ from pymatgen.entries.computed_entries import ComputedEntry, ComputedStructureEn
 from pymatgen.io.vasp import Incar, Kpoints, Poscar
 
 from emmet.core.common import convert_datetime
-from emmet.core.mpid import AlphaID, MPID
+from emmet.core.mpid import MPID, AlphaID
 from emmet.core.structure import StructureMetadata
+from emmet.core.trajectory import Trajectory as CoreTrajectory
 from emmet.core.utils import utcnow
 from emmet.core.vasp.calc_types import (
     CalcType,
@@ -35,6 +36,7 @@ from emmet.core.vasp.calc_types import (
 from emmet.core.vasp.calculation import (
     Calculation,
     CalculationInput,
+    CoreCalculationOutput,
     PotcarSpec,
     RunStatistics,
     VaspObject,
@@ -44,6 +46,7 @@ from emmet.core.vasp.utils import TASK_NAMES, discover_and_sort_vasp_files
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
+
     from typing_extensions import Self
 
 monty_decoder = MontyDecoder()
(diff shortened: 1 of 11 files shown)


esoteric-ephemera:
Will do! Yeah the `ProductionTaskDoc.output` should be `CalculationOutput` or a subset of its fields

tsmathis:
@esoteric-ephemera, this is ready for you to look at for field choices

(this is up to date w/ `main` so #1226 and #1230 should be accounted for)

tschaume:
re: question about `builder_meta`. I agree that removing all fields but `license` makes sense. Depending on how much it would break downstream, I'd even advocate for removing `builder_meta` in tasks entirely and make `license` a root-level field. `license` is not really a builder-related field even for downstream collections. I don't mind re-writing affected code in web and api if needed.

tsmathis:
top level license would be fine. we can think of alternatives, but we _do_ need some sort of explicit embedding for TOS, right?

emmet-core/emmet/core/vasp/validation_legacy.py
```diff
@@ -436,20 +463,32 @@ def _potcar_stats_check(
...
+            )
+    else:
+        if (outcar := calc[0]["output"]["outcar"]) and (
+            mag_info := outcar.get("magnetization", [])
+        ):
+            return any(
+                abs(mag_info[isite].get("tot", 0.0))
+                > abs(max_magmoms.get(site.label, np.inf))
+                for isite, site in enumerate(structure)
+            )
```
tschaume, line 490: The return values are the same. Could be updated to not repeat code.

emmet-core/emmet/core/tasks.py
```diff
@@ -243,147 +246,219 @@ def from_vasp_calc_docs(
...
-    task_label: str | None = Field(None, description="A description of the task")
+    @field_validator("batch_id", mode="before")
+    def validate_batch_id(cls, batch_id: str):
+        if batch_id:
+            invalid_chars = set(
+                char
+                for char in batch_id
+                if (not char.isalnum()) and (char not in {"-", "_"})
+            )
+            if len(invalid_chars) > 0:
```
tschaume, line 316: `if invalid_chars`?

emmet-core/emmet/core/tasks.py
```diff
@@ -243,147 +246,219 @@ def from_vasp_calc_docs(
...
+            props["structure"].append(ionic_step.structure)
+            props["energy"].append(ionic_step.e_0_energy)
+            for k in (
+                "e_fr_energy",
+                "e_wo_entrp",
+                "forces",
+                "stress",
+                "electronic_steps",
+            ):
+                props[k].append(getattr(ionic_step, k))
```
tschaume, line 407: What if `k` doesn't exist in `ionic_step`?

tschaume (changes requested):
Looks good! Only have minor/picky suggestions for your consideration.

tschaume:
> top level license would be fine. we can think of alternatives, but we _do_ need some sort of explicit embedding for TOS, right?

Good question. I'm not sure we need to embed (a link to the) TOS in the documents explicitly. Maybe a root-level `README/LICENSE/TOS` file is enough?

emmet-core/emmet/core/tasks.py
tsmathis: @esoteric-ephemera, I stole this from your trajectory class, thoughts?

looking back through logs when I was converting the 1.9 M tasks table to parquet and splitting the trajectories out I got a good number of:

```
2025-05-20 10:51:20,678 - tasks-trajectories-parquet - WARNING - Failed to build trajectory for mp-1263990, (alpha_id: aaactxva)
2025-05-20 10:51:20,678 - tasks-trajectories-parquet - WARNING - Exception: 'NoneType' object has no attribute 'is_ordered
 ```

try/except I had:
```python
try:
    traj = Trajectory._from_dict(props, task_type=tt, run_type=rt, identifier=task_id)
except Exception as e:
    logger.warning(f"Failed to build trajectory for {doc['task_id']}, (alpha_id: {task_id})")
    logger.warning(f"Exception: {e}")
else:
    _dict = traj.model_dump(mode="json")
    conv_data = traj.convergence_data
    for k, v in conv_data.items():
        _dict[f"{k}_convergence"] = v

    _dict["has_full_output"] = True

    trajs.append(_dict)
```

emmet-core/emmet/core/vasp/material.py
```diff
@@ -201,14 +202,34 @@ def _entry_eval(task: TaskDoc):
                 key=_entry_eval,
             )
 
-            if len(relevant_calcs) > 0:
+            if relevant_calcs:
```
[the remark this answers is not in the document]
tsmathis: this was just a clean up I saw. Only thing that matters for this branch is whether or not the `relevant_calcs` list is empty or not

emmet-core/emmet/core/vasp/validation_legacy.py
tsmathis: same comment as above re: scrapping the backwards compatibility branching -> would be back to one path here

tsmathis:
> Good question. I'm not sure we need to embed (a link to the) TOS in the documents explicitly. Maybe a root-level `README/LICENSE/TOS` file is enough?

Might be a bit tricky when adding in iceberg/delta on top of the parquet tables. I am not aware of the ability to add files into the root of a table's directory that won't be tracked/need to conform to the schema. Not sure if something comparable to dotfiles exists for those technologies. Will investigate

emmet-core/emmet/core/vasp/validation_legacy.py
```diff
@@ -195,20 +210,29 @@ def from_task_doc(
...
+                if calcs_reversed:
+                    if _magmom_check(
+                        calcs_reversed, structure, max_magmoms=max_magmoms
+                    ):
+                        reasons.append(DeprecationMessage.MAG)
+                else:
+                    if _magmom_check(
+                        task_doc.output, structure, max_magmoms=max_magmoms
+                    ):
+                        reasons.append(DeprecationMessage.MAG)
```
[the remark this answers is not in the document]
tschaume: I'm happy to defer to you and @esoteric-ephemera for when it's best to scrap. If you'd like to postpone that decision, the following code should do the same and be straightforward to edit when needed, right?

```python
if _magmom_check(
    calcs_reversed or task_doc.output, structure, max_magmoms=max_magmoms
):
    reasons.append(DeprecationMessage.MAG)
```

tsmathis:
> > top level license would be fine. we can think of alternatives, but we _do_ need some sort of explicit embedding for TOS, right?
> 
> Good question. I'm not sure we need to embed (a link to the) TOS in the documents explicitly. Maybe a root-level `README/LICENSE/TOS` file is enough?

@esoteric-ephemera , @tschaume 

d642c5c12 -> covers the meta change we discussed?

emmet-core/emmet/core/vasp/validation_legacy.py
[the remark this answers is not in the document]
tsmathis: 5f8a0be6c better?

emmet-core/emmet/core/tasks.py
tsmathis: yeah since you have a convenience function going there will you update your PR to incorporate your changes after I merge this into main?

esoteric-ephemera: Yep will do and will try to take care of the unrelated test that's now failing

PR #1246, opened 2025-06-10
Updating all type annotations to >3.10 conventions

tsmathis:
ripping the bandaid off -_-

making diffs easier to parse in #1243


emmet-api/emmet/api/routes/legacy/jcesr/query_operators.py
```diff
@@ -45,43 +45,53 @@ class MoleculeBaseQuery(QueryOperator):
 
     def query(
         self,
-        nelements_max: Optional[float] = Query(
+        nelements_max: float
+        | None = Query(
```
tschaume, line 49: Doesn't make a difference but it's a weird linebreak. I'd expect everything up to `Query(` to be on the same line

tschaume (approved):
Looks good! The linebreaks for the `| None` are a little off I think. But that's nothing major.

emmet-api/emmet/api/routes/legacy/jcesr/query_operators.py
tsmathis: I found that weird too, my local formatter had that originally knocked down to one line, `pre-commit` split that up. Seems to only happen on annotations of function args.

tschaume: Ah good catch! Yes, we can fold the bump into this PR.

Issue #1249, opened 2025-06-13
[Bug]: Spin-Orbit Coupling in VASP Workflows


zhenyaofang:
### Code snippet

```python

```

### What happened?

Hi,

I'm reaching out about the failure of validating magnetic moments in spin-orbit coupling (SOC) calculations in VASP. 

I was testing a workflow in `Atomate2` with SOC enabled, and I ran into a `pydantic` validation error related to the magnetic moments. In SOC calculations, magnetic moments are given as a list of three float numbers, but it seems that `pydantic` is expecting a single float number instead.

I was wondering if there is any workaround for this issue. Any guidance would be much appreciated.

Thanks very much.

### Version

0.84.6

### Which OS?

- [ ] MacOS
- [ ] Windows
- [x] Linux

### Log output

```shell
pydantic_core._pydantic_core.ValidationError: 4 validation errors for InputDoc
magnetic_moments.0
  Input should be a valid number [type=float_type, input_value=[0.0, 0.0, 0.0], input_type=list]
    For further information visit https://errors.pydantic.dev/2.9/v/float_type
```

tsmathis:
The current type annotation is `list[float]`
https://github.com/materialsproject/emmet/blob/076a08618324d6b4b07ef983484b6a4b14a8429e/emmet-core/emmet/core/tasks.py#L197

So for the SOC calculations, is the input an array of float arrays?

Can you link me to the entry point in atomate2 where enabling SOC happens?

esoteric-ephemera:
@tsmathis it's a run type we don't use at MP but is a valid possibility (for SOC / noncolinear jobs in general, the magnetic_moments will be `list[Vector3D]`)

Probably better to just make a separate field for this, so we avoid mixed types

esoteric-ephemera:
Hey @zhenyaofang  can you try `pip install emmet-core==0.84.8` and see if this is fixed?

zhenyaofang:
Hi @esoteric-ephemera,

Yes, it's fixed now. Thanks so much for your help.

Issue #1248, opened 2025-06-13
[Bug]: _get_state in core/tasks.py


ThomasWarford:
### Code snippet

```python
from atomate2.vasp.jobs.core import RelaxMaker
from atomate2.vasp.flows.matpes import MatPesStaticFlowMaker
from jobflow import run_locally
from pymatgen.core import Structure

# construct an FCC silicon structure
si_structure = Structure(
    lattice=[[0, 2.73, 2.73], [2.73, 0, 2.73], [2.73, 2.73, 0]],
    species=["Si", "Si"],
    coords=[[0, 0, 0], [0.25, 0.25, 0.25]],
)

# make a relax job to optimise the structure
relax_job = MatPesStaticFlowMaker().make(si_structure)

# run the job
run_locally(relax_job, create_folders=True)




https://github.com/materialsproject/emmet/blob/076a08618324d6b4b07ef983484b6a4b14a8429e/emmet-core/emmet/core/tasks.py#L1044
```

### What happened?

I ran the above example script with atomate2 (MatpesStaticFlowMaker has replaced the original RelaxMaker) from https://materialsproject.github.io/atomate2/user/install.html.

I get the below error, despite the VASP output files looking correct.

I tracked the error down to this line: https://github.com/materialsproject/emmet/blob/076a08618324d6b4b07ef983484b6a4b14a8429e/emmet-core/emmet/core/tasks.py#L1049

```
def _get_state(calcs_reversed: list[Calculation], analysis: AnalysisDoc) -> TaskState:
    """Get state from calculation documents and relaxation analysis."""
    all_calcs_completed = all(
        c.has_vasp_completed == TaskState.SUCCESS for c in calcs_reversed
    )
    if analysis.errors and len(analysis.errors) == 0 and all_calcs_completed:
        return TaskState.SUCCESS  # type: ignore
    return TaskState.FAILED  # type: ignore
```

analysis.errors is an empty list (which is False in python), so the output will be TaskState.FAILED despite `len(analysis.errors)` being zero. 

As long as `analysis.errors` is a list this will return TaskState.FAILED.

I might be missing something here.

### Version

emmet-core                0.84.7rc4

### Which OS?

- [ ] MacOS
- [ ] Windows
- [x] Linux

### Log output

```shell
2025-06-13 15:47:22,028 INFO MatPES GGA static failed with exception:
Traceback (most recent call last):
[...]

PR #1250, opened 2025-06-13
Fix TaskDoc bugs in latest emmet-core

esoteric-ephemera:
- #1248 : better check on when `AnalysisDoc.errors` is non-populated
- #1249 : temporary fix for non-colinear magnetic moments while #1226 is pending


tsmathis:
yep all good, since this is for bug fixes lets just put the patch label on and let the actions handle it

PR #1252, opened 2025-06-16
Construct trajectory from MLMD data when present

mcgalcode:
WIP

### DIFF
--- a/emmet-core/emmet/core/vasp/calculation.py
+++ b/emmet-core/emmet/core/vasp/calculation.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import logging
 import os
+import copy
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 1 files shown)


gpetretto:
Hi @mcgalcode @esoteric-ephemera, it is nice to see that there is much going on to improve the support for trajectories in MD simulations. I think this is a broad topic and maybe it would be better to limit the discussions in a few places. However I have at least one comment that is specific to this PR. As I have partially mentioned in #872 and https://github.com/materialsproject/atomate2/issues/515, my experience is that it is relatively easy to generate very large trajectories (in the order of the 100000 steps). In my case it was already a challange to parse the outputs from the vasprun.xml with pymatgen, both in terms of time and memory. In this respect I would suggest to avoid making a full copy of all the `md_data` and then discard the structures, because I expect that for a long trajectory the deepcopy may result in a considerable overhead.

esoteric-ephemera:
@gpetretto for these really long MD runs, have you tried parsing a trajectory from the vaspout.h5? The Vaspout parser in pymatgen should only load objects as needed, and you could in principle just load in the ionic step data

Asking primarily because it might be better to switch to vaspout.h5 parsing overall (more calculation detail, faster parsing)

esoteric-ephemera:
Yeah the long term plan is to use vaspout.h5 parsing if possible, if only because this is what VASP appears to be developing (new features are only added to vaspout.h5, not vasprun.xml)

The pyarrow/parquet `emmet.core.trajectory.Trajectory` class should natively support MD runs (it's intended to replace pymatgen's Trajectory) but there isn't a constructor directly from XDATCAR or vaspout.h5 to it. Happy to add that if it would be useful!

esoteric-ephemera:
@mcgalcode we've significantly changed how trajectories get parsed from vasprun.xml since you opened this: There's a pydantic-based model in emmet core for this. By default, you should still get a pymatgen Trajectory unless you opt into the [emmet models](https://github.com/materialsproject/emmet/blob/74194bbf8c7b32ce15141bb1c9ee0527a0fc6c45/emmet-core/emmet/core/settings.py#L176) 

I integrated your changes  in [6a48ce1](https://github.com/materialsproject/emmet/pull/1380/commits/6a48ce1832b37aa0bc39e5d9296cddcea9b3f367) to prepare for our next deployment

If you end up needing other changes beyond those in [#1380](https://github.com/materialsproject/emmet/pull/1380), feel free to re-open this!

PR #1223, opened 2025-04-14
Update VASP file discovery

esoteric-ephemera:
Adds optionally-recursive VASP file discovery with enough flexibility to support current use in `TaskDoc.from_directory`, the new [emmet-archival](https://github.com/materialsproject/emmet/pull/1190) package, and the new [emmet-cli](https://github.com/materialsproject/emmet/pull/1211) - copying @kbuma

Defines categories of acceptable VASP I/O and workflow files for file discovery. Allows for arbitrary directory structure in finding files, and specification of directory depth

Misc / housekeeping:
- Remove `PymatgenTest` from test files, as this is [deprecated](https://github.com/materialsproject/pymatgen/pull/4212)

PR #1253, opened 2025-06-16
Bug fixes, continuing PEP585 support, PropertyDoc schema

esoteric-ephemera:
## Bug fixes:
- Ensure OSZICAR is parsed by file discovery, was unintentionally removed in #1223 (only affects MD runs by parsing temperature) - thanks @mcgalcode for the fix!
- Allow for TaskDoc to still parse VASP calcs with missing projected DOS data

## PEP 585
- Remove `typing.Set` in favor of `set`, and migrate objects like typing.Sequence to collections.abc.Sequence per the [migration guide](https://peps.python.org/pep-0585/#implementation)

## Schema Changes:
- Per internal discussion: made `structure` an optional top-level on `PropertyDoc` that is instantiated with `from_structure`. However, this field is by default both excluded from the `model_dump` and the JSON schema
  - Any class that already had a top-level structure, like `SummaryDoc`, overrides this behavior and includes the field in `model_dump` / JSON schema


tsmathis:
Okay guys, what is going on between this, #1252, and #1230

@mcgalcode your submission is for MLMD trajectory parsing, but has a patch for oszicar discovery?

@esoteric-ephemera you cherry-picked (?) [9c636d6](https://github.com/materialsproject/emmet/pull/1253/commits/9c636d612ce7260075b1ca14aa321bd129543f62) which only adds the "oszicar" to the vasp discovery? But not the relative file path patch for the parsing? I thought all of the oszicar stuff was in #1230. Is the file discovery the only missing piece?
You also have DOS parsing here?

esoteric-ephemera:
This is mostly patch / bug fixes for TaskDoc, related to changes in 1223. I left out the `oszicar_file` from the file discovery unintentionally, as it's not used in production data. Cherry picked Max's commit to give credit there

The DOS parsing changes here allow TaskDoc to still parse when complete DOS info is missing, before it would throw an exception and fail

Max's 1252 is mostly expanding the scope of TaskDoc to handle different run / task types (an MLMD run, something that's not in production data yet)

PR #1255, opened 2025-07-07
Avoid extrapolation in NEB fitting

gpetretto:
While calculating the barrier for a NEB calculation based on a spline fit, the roots of the derivative may also be in a region outside the [0,1] interval and this may result in a much high barrier. For example one of the fits I got had this shape and the barrier value came from the outer peaks, while the NEB fitted data is of course only comprised in the [0,1] region: 
fit

To fix this it seems enough to specify to avoid extrapolation.


esoteric-ephemera:
Great thanks for the bug fix!

PR #1256, opened 2025-07-08
NEB task doc refactoring

esoteric-ephemera:
Related to changes in underlying NEB workflows in atomate2 - these changes are for better structuring of NEB data, especially the final analyzed reaction path data:
- New `NebIntermediateImagesDoc` which represents the VASP NEB calculation of non-endpoint images 
- Much lighter `NebTaskDoc` which only contains high level reaction pathway info (structures, initial structures, energies, analyzed barriers)


esoteric-ephemera:
@tsmathis and @tschaume : these are parallel updates only to the Approx / NEB schemas to keep workflow development moving. Going to merge in so that this comes out in the next rc with AlphaID

PR #1259, opened 2025-07-15
Similarity builder

esoteric-ephemera:
### Summary
- Recreates the functions used to produce [materials similarity scores](https://doi.org/10.1039/C9RA07755C)
- Allows for rebuilding collection of similarity scores, which has not been updated
- Starts to build out ML-based similarity metrics using the M3GNet formation energy model
- Vendi scoring for diversity of dataset

### Motivation
- Integrate more ML-based metrics in MP + leverage vector search in structure data
- Use this to make an app that recommends structures in MP by similarity to a user-input structure

### To Do's:
- Incorporate other ML models (M3GNet interatomic potential + MatPES, MACE-MP-0). Issue is combining site embeddings / descriptors into a structural feature descriptor


esoteric-ephemera:
@tsmathis @tschaume I've validated that the builder is working now by building out a similarity collection for all 210k materials in summary. Shared with you on CFS (586 MB): `/global/cfs/cdirs/matgen/esoteric/mp_similarity.json.gz`

This took about 2 hours (combined creating the feature vectors and getting only the top 104 most similar materials to any given material), and is run like so:

```py

import numpy as np
from pathlib import Path

from pymatgen.core import Structure

import pandas as pd

from mp_api.client import MPRester
from emmet.core.similarity import CrystalNNSimilarity
from monty.serialization import dumpfn, loadfn

if (mp_struct_file := Path(".mp_structures.json.gz")).exists():
    mats = loadfn(mp_struct_file,cls=None)
else:
    with MPRester(use_document_model=False,monty_decode=False) as mpr:
        mats = mpr.materials.summary.search(fields=["material_id","structure"])

    dumpfn([{k : doc[k] for k in ("material_id","structure")} for doc in mats], mp_struct_file)

similarity_docs = CrystalNNSimilarity().build_similarity_collection_from_structures(
    {doc["material_id"]: Structure.from_dict(doc["structure"]) for doc in mats},
    num_procs=num_procs,
    num_top=104
)
```

(the `num_top` / number of most similar materials is arbitrary, just set it to 104 see how unfavorably this scales)

PR #1261, opened 2025-07-25
address robocrys search issues

kbuma:
This fix addresses two robocrys text search issues:

1. The search results returned through the API client would occasionally contain duplicates or missing items. (reported in https://matsci.org/t/mp-api-robocrys-keyword-search-misses-materials/64544) The implementation of the API client paginator relies on retrieving the same set of results in the same order every time the same query is issued. The aggregation pipeline set up to execute the search for the robocrys text search had an issue where that could lead to that not being the case. The sorting function used only sorted on the search score. If multiple results had the same search score in this case there would be no guarantee in how they would be ordered in respect to each other. The fix adds a secondary sort key that is used in the case of ties which ensures a unique order every time.
2. During the investigation of the above, it was discovered that the current aggregation pipeline would break when the number of matching result was greater than about 30k. The implementation made use of a facet stage which can utilize a lot of memory and mongo has a strict limit on the size it allows per aggregation pipeline. The pipeline has been optimized such that it does not require a facet stage and should now be able to handle any size result set.

## Contributor Checklist

- [x] I have verified that the aggregation pipeline produces the correct results in a performant manner
- [ ] I have run the tests locally and they passed.
- [ ] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR
- [ ] I have verified that the client API works properly with this use case.

### DIFF
--- a/emmet-api/Dockerfile
+++ b/emmet-api/Dockerfile
@@ -17,11 +17,11 @@ RUN SETUPTOOLS_SCM_PRETEND_VERSION=${API_VERSION} pip install $PIP_FLAGS --no-de
 RUN wget -q https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh && \
   chmod +x wait-for-it.sh && mv wait-for-it.sh /root/.local/bin/
 
-#FROM base
-#COPY --from=builder /root/.local/lib/python3.11/site-packages /root/.local/lib/python3.11/site-packages
-#COPY --from=builder /root/.local/bin /root/.local/bin
-#COPY --from=builder /usr/lib/x86_64-linux-gnu/libsnappy* /usr/lib/x86_64-linux-gnu/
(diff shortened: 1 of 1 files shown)


emmet-api/emmet/api/routes/materials/robocrys/query_operators.py
```diff
@@ -32,43 +32,26 @@ def query(
...
-            {"$unwind": "$total_doc"},
-            {
-                "$replaceRoot": {
-                    "newRoot": {
-                        "$mergeObjects": ["$results", {"total_doc": "$total_doc.count"}]
-                    }
+                "$project": {
+                    "_id": 0,
+                    "meta": "$$SEARCH_META",
+                    "material_id": 1,
```
tschaume, line 45: Looks like there was a typo in the original pipeline. We should keep `last_updated` as a projected field.

PR #1278, opened 2025-08-25
Use emmet models for VASP objects instead of pymatgen objects

esoteric-ephemera:
Converts `TaskDoc` and related parsing/retrieval in emmet-api to using more efficient emmet models for data objects:
- `Trajectory`
- `ElectronicDos` instead of `Dos` and `CompleteDos`
- `ElectronicBS` instead of `BandStructure` and `BandStructureSymmLine`

Misc changes:
- Allows for converting an `ElectronicBS` object to any `BandStructure`-compatible object, such as `BandStructureSymmLine`
- Drop py==3.10 support consistent with `numpy` [dropping support last year](https://github.com/numpy/numpy/commit/688669f81c13fc98671db97a464a57604d4e1a0b)


esoteric-ephemera:
@tsmathis can you take a look at the `ProjectedDos` and `ProjectedBS` objects and let me know if that's more along the lines of the columnar organization we were discussing?

tsmathis:
Yeah those are looking good for data layout. The projection models should include an `identifier` field, no?

And one thought: For the "complete" models (`ElectronicBS`, `ElectronicDos`) should the `projections` and `projected_densities` be something like a private attr or a computed field? 

Or should these be treated like the outputs of some parsing code, and at runtime a choice should be made if documents should be split (based on however the consumer is storing data)? -> If this is the case then the identifier could be split off at runtime as well

esoteric-ephemera:
> Or should these be treated like the outputs of some parsing code, and at runtime a choice should be made if documents should be split (based on however the consumer is storing data)?

I was thinking more of the run time solution / only writing to a table. But yeah will add the identifier field, thanks for reminding me

tschaume (dismissed):

emmet-api/emmet/api/routes/materials/tasks/query_operators.py
```diff
@@ -96,15 +96,16 @@ def query(
...
         """
 
         d = [
             {
                 "task_id": doc["task_id"],
-                "trajectories": jsanitize(
-                    calcs_reversed_to_trajectory(doc["calcs_reversed"])
-                ),
+                "trajectories": [
+                    traj.model_dump(mode="json")
```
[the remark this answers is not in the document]
esoteric-ephemera: In `model_dump`? Not by default - pydantic seem [opposed](https://github.com/pydantic/pydantic/discussions/6388#discussioncomment-13944196) to the idea of using `orjson` over their own `jiter` parser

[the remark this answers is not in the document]
tschaume: oh nice! Sounds it might be a good idea to use `jiter` across the board then and save ourselves the `orjson` dependency. @tsmathis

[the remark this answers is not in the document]
esoteric-ephemera: @tschaume don't think we can remove the `orjson` dependence since pymatgen [requires it](https://github.com/materialsproject/pymatgen/blob/ab34799d8ab5dee80756489cf2ca28a97de78121/pyproject.toml#L66)

I can look around and see if there are spots where we can replace `json` with `orjson` so we at least make use of the dependence

emmet-core/emmet/core/band_theory.py
```diff
@@ -63,6 +63,58 @@ def reciprocal_lattice_deserializer(cls, reciprocal_lattice):
...
+    tensor where:
+        - Index 1: band index
+        - Index 2: k-point index
+        - Index 3: site index
+        - Index 4: index for the orbital character, ordered as
+            s, py, pz, px, dxy, dyz, dz2, dxz, dx2-y2,...
+    """
+
+    spin_up: list[float] | None = Field(
+        None, description="The flattened spin-up band projetions."
```
tschaume, line 81: typo "proje(c)tions"

emmet-core/emmet/core/band_theory.py
```diff
@@ -63,6 +63,58 @@ def reciprocal_lattice_deserializer(cls, reciprocal_lattice):
...
+    )
+    spin_down: list[float] | None = Field(
+        None, description="The flattened spin-down band projetions."
```
tschaume, line 84: typo

emmet-core/emmet/core/band_theory.py
```diff
@@ -163,91 +205,134 @@ def to_pmg(
...
+            for i in range(len(self.spin_up_densities or []))
+        ]
 
-    orbital: str | None = Field(
-        None, description="The orbital character of this DOS, if applicable."
-    )
+    @classmethod
+    def _from_list_of_dict(cls, lod: list[dict[str, str | int | list[float]]]) -> Self:
+        """Possible deserialization route."""
+        return cls(**{k: [entry.get(k) for entry in lod] for k in cls.model_fields})  # type: ignore[arg-type]
```
tschaume, line 242: Should the list entry be `None` explicitly when `k not in entry`? I.e.

```python
[entry.get(k) for entry in lod]
```

vs

```python
[entry[k] for entry in lod if k in entry]
```

emmet-core/emmet/core/band_theory.py
```diff
@@ -163,91 +205,134 @@ def to_pmg(
...
+    def to_pmg_like(
+        self, structure: StructureType
+    ) -> dict[PeriodicSite, dict[Orbital, dict[Spin, np.ndarray]]]:
+        """Construct a pymatgen-like representation of the projected DOS."""
+        pdos: dict[PeriodicSite, dict[Orbital, dict[Spin, np.ndarray]]] = {}
+        for i, dens in enumerate(self.spin_up_densities or []):
+            if (site := structure[self.site_index[i]]) not in pdos:  # type: ignore[index]
+                pdos[site] = {}
+            if (orb := Orbital[self.orbital[i]]) not in pdos[site]:  # type: ignore[index,misc]
+                pdos[site][orb] = {}
```
tschaume, line 287: Could this be omitted/shortened by using `defaultdict(dict)`?

emmet-core/emmet/core/band_theory.py
```diff
@@ -163,91 +205,134 @@ def to_pmg(
...
 
     @property
     def _available_spins(self) -> list[Spin]:
         """Retrive spin indices of non-null spin data."""
-        return [
-            spin
-            for spin in Spin
-            if getattr(self, f"spin_{spin.name}_densities", None) is not None
-        ]
+        return [spin for spin in Spin if getattr(self, f"spin_{spin.name}_densities")]
```
tschaume, line 320: raise `AttributeError` or return `default=None` in `getattr`?

emmet-core/emmet/core/band_theory.py
```diff
@@ -265,36 +350,16 @@ def to_pmg(self) -> Dos | CompleteDos:
...
             )
 
-        densities: dict[Spin, np.ndarray] = {}
-        for spin in self._available_spins:
-            if (
-                _dens := getattr(self, f"spin_{spin.name}_densities", None)
-            ) is not None:
-                densities[spin] = np.array(_dens)
+        densities: dict[Spin, np.ndarray] = {
+            spin: np.array(getattr(self, f"spin_{spin.name}_densities"))
```
tschaume, line 354: `getattr` default arg missing?

emmet-core/emmet/core/band_theory.py
tschaume: @esoteric-ephemera @tsmathis Thanks for clarifying! Either solution is good I think. Whatever you think is best going forward.

emmet-core/emmet/core/band_theory.py
tschaume: Thanks! No need to rewrite when the model fields are guaranteed to exist.

esoteric-ephemera:
Also extending this PR to flesh out the EOS schemas in preparation for replacing these data points (only 232 in MP currently). To migrate to the new schema, it takes maybe 2 sec with the current data:

```py
from emmet.core.eos import EOSDoc
from mp_api.client import MPRester

with MPRester(use_document_model=False) as mpr:
    mats_w_eos = mpr.materials.summary.search(has_props=["eos"],)
    eos = mpr.materials.eos.search()
    
mpid_to_struct = {
    doc["material_id"] : doc["structure"] for doc in mats_w_eos
}
eos_docs = {
    doc["material_id"] : EOSDoc.from_ev_data(
        mpid_to_struct[doc["material_id"]],
        doc["volumes"],
        doc["energies"],
    )
    for doc in eos
}
```

emmet-archival/emmet/archival/core.py
```diff
@@ -0,0 +1,336 @@
...
+    """
+    if not path.is_relative_to(parent):
+        raise ValueError(f"Path {path} is not relative to {parent}")
+
+    if path == parent:
+        return Path("")
+
+    for p in path.parents:
+        if p == parent:
+            break
```
tschaume: Is there a scenario here where an `else` would be needed for the for-loop?

PR #1289, opened 2025-09-09
Fix the parsing of hint parameters for pymongo

kbuma:
Fix the parsing of hint parameters for pymongo. Pymongo sends a hint=null if hint is passed in as None rather than disregarding it.

## Contributor Checklist

- [ ] I have run the tests locally and they passed.


tschaume (approved):
Looks good! Thanks! I only cleaned up a little by adding a `self.get_search_kwargs()` that can be used for both `count_documents()` and `aggregate()`

tschaume (approved):

PR #1328, opened 2025-10-17
emmet-core 0.86.0.rc1

esoteric-ephemera:
Changes building on `emmet-core==0.86.0rc0` to allow for compatibility with the API client. A lot of these are issues with the data not actually conforming to the schema defined in emmet. Somehow these discrepancies weren't caught prior to now.

Summary:
- EOS: the `EOSDoc.eos` field was given as a single `EOSType` entry, however in the DB it is a dict mapping an EOS model (which sometimes includes typos) to `EOSType`. Adjusted to the `dict` to avoid data loss
- EOS: Added enum with typo handling to structure EOS data
- Provenance: schematize `history.description` to accommodate data types currently in database. Add migration constructor method to provenance and SNL docs

Note that some of the EOS changes will get superseded by [#1278](https://github.com/materialsproject/emmet/pull/1278)

### DIFF
--- a/emmet-api/emmet/api/routes/materials/materials/query_operators.py
+++ b/emmet-api/emmet/api/routes/materials/materials/query_operators.py
@@ -474,6 +474,9 @@ class LicenseQuery(QueryOperator):
 class BatchIdQuery(QueryOperator):
     """Method to generate a query on batch_id"""
 
+    def __init__(self, field="builder_meta.batch_id"):
+        self._field = field
+
     def query(
         self,
         batch_id: str | None = Query(
@@ -502,7 +505,7 @@ class BatchIdQuery(QueryOperator):
             )
 
         crit = {}  # type: dict
-        k = "builder_meta.batch_id"
+        k = self._field
         if batch_id:
             crit[k] = batch_id
         elif batch_id_not_eq:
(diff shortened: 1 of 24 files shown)


emmet-core/emmet/core/provenance.py
```diff
@@ -218,3 +277,124 @@ def from_SNLs(
...
+            descs, init_args, depth + 1, remark=f"init_args-depth-{depth}"
+        )
+
+    # always append current entry even if there were nested fields
+    orig_remark = entry.pop("remark", None)
+    if not remark:
+        remark = orig_remark
+
+    for k in ("class", "module"):
+        entry[f"monty_{k}"] = entry.pop(f"@{k}", None)
```
tsmathis: not a change request, but an option for handling the @module and @class fields for MSONables since those are invalid variable/field names. The alternate constructor for TypedDict allows for arbitrary strings: [structure_adapted.py](https://github.com/materialsproject/emmet/blob/aaaa087d94b36eec10111c26947f01f0b12e1fc9/emmet-core/emmet/core/types/pymatgen_types/structure_adapter.py#L15) - would remove the need for these types of monty specific loops.

And if you want to stick in pydantic land you can use a `validation_alias` on the `monty_*` fields [field aliases](https://docs.pydantic.dev/latest/concepts/fields/#field-aliases):
```python
...
    monty_class: str | None = Field(None, validation_alias="@class")
    monty_module: str | None = Field(None, validation_alias="@module")
...
```
then the logic for this loop is baked into the model

esoteric-ephemera: Field route sounds good since I'm not sure why these get populated as top-levels in the `description` - guessing some descriptions are just `Structure().as_dict()`??

PR #1330, opened 2025-10-23
add BatchIDQuery to tasks_resource

tsmathis:
related to https://github.com/materialsproject/api/pull/1023

@tschaume does this accomplish what I wanted to get for the client? Mainly unsure about the `batch_id_field` param in `BatchIdQuery`


tsmathis:
gotcha, wasn't sure if adding an `__init__` would mess things up.

how about 5fad69570de92c7c66fb037d72f852cd7a4f7993 then?

PR #1339, opened 2025-11-06
emmet 0.86.0: sync develop with main

esoteric-ephemera:
@tschaume for review once tests pass. Will need to sync main with develop right after


tschaume (approved):
Looks good! Thanks!

PR #1342, opened 2025-11-06
emmet 0.86.0

esoteric-ephemera:


esoteric-ephemera:
@tschaume one of the archival tests I added isn't super reliable (does byte literal comparison), removing that in #1343 and then this should be gtg

tsmathis (approved):
nice, thanks for getting the sync put together

PR #1345, opened 2025-11-08
Various patches for atomate2/workflow support

esoteric-ephemera:
Bug fixes / patches for atomate2 support. Will hopefully define emmet-core==0.86.1rc0:

- Missed this, but pymatgen can write a `POTCAR.spec` file that is a newline-delimited list of the TITEL fields. Right now, the auto-parsing of this breaks because we assume JSON format. **Accommodate both formats in parsing, with preference to the JSON format.**
- There are a lot of "hidden" `pymatgen.io.vasp.outputs.Outcar` attributes that ***only*** get set under specific circumstances. Updating the typed `Outcar` dict to accommodate these
- Fixing docstr's that reference relative paths when they are actually absolute paths
- Locpot parsing into averaged arrays
- Support for old double-relax style calcs in TaskDoc

### DIFF
--- a/emmet-archival/tests/vasp/test_volumetric.py
+++ b/emmet-archival/tests/vasp/test_volumetric.py
@@ -61,7 +61,6 @@ def test_volumetric_archive():
 
     # Note that pymatgen doesn't parse augmentation charge data
     # cleanly - the data won't be equal here but the keys are.
-    print(chg.data_aug)
     assert all(
         k.value in chg.data_aug
         and (chg_arch.data_aug[i] is not None if chg.data_aug[k] else True)


esoteric-ephemera:
@tschaume and @tsmathis these changes are all additive/tangential to support atomate2, once tests pass I'd like to put out an rc for this so I can also mint a new atomate2 release

emmet-core/emmet/core/vasp/calculation.py
```diff
@@ -1016,14 +1021,14 @@ def from_vasp_files(
         locpot = None
         if average_locpot:
             if VaspObject.LOCPOT in vasp_objects:
-                locpot = vasp_objects[VaspObject.LOCPOT]  # type: ignore
+                locpot = vasp_objects[VaspObject.LOCPOT].to_pmg()  # type: ignore
```
tsmathis: Just a question for my reference later, the locpot here will be a `ChgcarLike` from `emmet.core.vasp.models`?

PR #1347, opened 2025-11-11
Crystalline similarity vector search

esoteric-ephemera:
Use Atlas vector search to allow users to query materials by similarity, using a `CrystalNN`-derived feature vector

To eventually be expanded to ML-based methods, see TODO's therein

### DIFF
--- a/emmet-api/material_resources.py
+++ b/emmet-api/material_resources.py
@@ -46,7 +46,7 @@ if db_uri:
     )
     task_store = CollectionWithKey(tasks_db["tasks"], "task_id")
     eos_store = CollectionWithKey(core_db["eos"], "task_id")
-    similarity_store = CollectionWithKey(core_db["similarity_new"])
+    similarity_store = CollectionWithKey(core_db["similarity_crystalnn_2025_11_05"])
     xas_store = CollectionWithKey(core_db["xas"], "spectrum_id")
     gb_store = CollectionWithKey(core_db["grain_boundaries"], "task_id")
     fermi_store = CollectionWithKey(core_db["fermi_surface"], "task_id")
--- a/emmet-core/tests/test_model_fields.py
+++ b/emmet-core/tests/test_model_fields.py
@@ -905,6 +905,7 @@ ref_model_fields = {
         "density_atomic",
         "symmetry",
         "material_id",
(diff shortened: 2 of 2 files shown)


emmet-api/tests/materials/similarity/test_query_operators.py
```diff
@@ -0,0 +1,50 @@
...
+from emmet.core.similarity import SimilarityMethod
+from emmet.api.routes.materials.similarity.query_operators import (
+    SimilarityFeatureVectorQuery,
+    SIM_METHOD_TO_FEAT_VEC_LENGTH,
+)
+import numpy as np
+
+
+@pytest.mark.parametrize(
+    "method", [None, "CrystalNN", "CRYSTALNN", SimilarityMethod.CRYSTALNN]
```
tsmathis: just need a `.value` on `SimilarityMethod.CRYSTALNN` to get your test to run right?

tsmathis: ah sorry, wrong. Should be failing here?
https://github.com/materialsproject/emmet/blob/b583a48c44d76a151449a0d25b29eb339c71a2f6/emmet-api/emmet/api/routes/materials/similarity/query_operators.py#L54-L56

emmet-api/emmet/api/routes/materials/similarity/query_operators.py
```diff
@@ -0,0 +1,110 @@
...
+                detail=f"Invalid feature vector for method {method.value}: "
+                f"should be a list of {ref_fv_len} floats.",
+            )
+
+        index_name = "similarity_feature_vector"
+        # because MongoDB does not permit renaming indexes,
+        # and I was not forward thinking in naming it.
+        # TODO: homogenize once we have other data built out
+        if method != SimilarityMethod.CRYSTALNN:
+            index_name += f"_{method.value.lower()}"
```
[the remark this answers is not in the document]
esoteric-ephemera: Right now, each vector search index needs to explicitly set the length of the vector to search, so we'd need a separate index for each similarity measure

I just wasn't forward thinking in naming this, if we expand out the similarity measures then I can rebuild the index with a new name

PR #1349, opened 2025-11-20
add path_convention determination to ElectronicBS

tsmathis:
Would be useful to add the `path_convention` to the emmet band structure model.

@esoteric-ephemera, does the implementation of `obtain_path_type` even make sense to you? Originated in the es builder.

Trying the function on just the `structure` and `orig_inputs.kpoints` for the subset of tasks I tried failed for everything but `latimer_munro`. Figured using the actual band structure object must be more reliable?

### DIFF
--- a/emmet-core/tests/test_band_theory.py
+++ b/emmet-core/tests/test_band_theory.py
@@ -2,6 +2,8 @@ import json
 
 import numpy as np
 import pytest
+
+from monty.serialization import loadfn
 from pymatgen.electronic_structure.core import Spin
 from pymatgen.electronic_structure.dos import CompleteDos, Dos
 from pymatgen.io.vasp import Vasprun
@@ -12,7 +14,10 @@ from emmet.core.band_theory import (
     ProjectedBS,
     ElectronicDos,
     ProjectedDos,
+    obtain_path_type,
+    get_path_from_bandstructure,
 )
+from emmet.core.electronic_structure import BSPathType
 from emmet.core.testing_utils import DataArchive
 
 if ARROW_COMPATIBLE:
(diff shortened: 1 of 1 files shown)


emmet-core/emmet/core/band_theory.py
```diff
@@ -367,3 +385,49 @@ def to_pmg(self) -> Dos | CompleteDos:
...
+def obtain_path_type(
+    labels_dict,
+    structure,
+    symprec=SETTINGS.SYMPREC,
+    angle_tolerance=SETTINGS.ANGLE_TOL,
+    atol=1e-5,
+):
+    bs_type = None
+
+    if any([label.islower() for label in labels_dict]):
```
esoteric-ephemera: This is interesting because it just basically checks to see if the path order {$\Gamma$, X, K, ...} is specified as lowercase letters, and then assumes latimer_munro. That's odd because the actual [kpath classes in pymatgen](https://github.com/materialsproject/pymatgen/blob/b985b94eb01b84a0c72e8036b61f9928545a8013/src/pymatgen/symmetry/kpath.py#L1008) **all** return lowercase labels. Maybe this was guaranteed based on builder/atomate behavior?

tsmathis: yeah the entire thing seemed a bit dubious to me, but I am not close enough to that code to have an idea of what should or should not make sense. 

But just as an example out of 1000 tasks (not the actual band structure), the `obtain_path_type` function only identified <200 `latimer_munro` entries, the rest were null. Heres a few `path_type`, `label_dict`s as examples:

```
{}
None
{'r': [0, 0, 0], 'q': [0, 0, 0], 'a': [0, 0, 0], 'Γ': [0, 0, 0], 'q_{1}': [0, 0, 0], 'd': [0, 0, 0]}
latimer_munro
{'\\Gamma': [0, 0, 0], 'X': [0, 0, 0], 'W': [0, 0, 0], 'K': [0, 0, 0], 'L': [0, 0, 0], 'U': [0, 0, 0]}
None
{'\\Gamma': [0, 0, 0], 'X': [0, 0, 0], 'W': [0, 0, 0], 'K': [0, 0, 0], 'L': [0, 0, 0], 'U': [0, 0, 0]}
None
{'GAMMA': [0, 0, 0], 'X': [0, 0, 0], 'U': [0, 0, 0], 'K': [0, 0, 0], 'L': [0, 0, 0], 'W': [0, 0, 0]}
```

esoteric-ephemera (approved):

PR #1350, opened 2025-11-21
Change rotation_number type from float to int

Andrew-S-Rosen:
Simple typing fix to update `CrystalSystem.rotation_number` to `int`.

### DIFF
--- a/emmet-core/emmet/core/symmetry.py
+++ b/emmet-core/emmet/core/symmetry.py
@@ -74,7 +74,7 @@ class PointGroupData(BaseModel):
     )
 
     @field_validator("rotation_number", mode="before")
-    def remove_inf(cls, v : Any) -> int | None:
+    def remove_inf(cls, v: Any) -> int | None:
         if v is not None:
             v = None if isinf(float(v)) else int(v)
         return v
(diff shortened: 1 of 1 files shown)


emmet-core/emmet/core/symmetry.py
```diff
@@ -44,7 +44,7 @@ class PointGroupData(BaseModel):
         None, title="Point Group Symbol", description="The point group for the lattice"
     )
 
-    rotation_number: float | None = Field(
+    rotation_number: int | None = Field(
```
esoteric-ephemera: For sure the molecular symmetry number should be an `int`, but the document model tries to encode the infinite symmetry order of an isolated atom as `float("inf")`.  Changing this to `int | float | None` should be a good middle ground

esoteric-ephemera: There's no int equivalent of `float("inf")` so my vote would be to just leave it as `float` in that case

esoteric-ephemera:
@Andrew-S-Rosen and @tsmathis: looking through the molecules database, I think we already removed the `float("inf")` rotation numbers - they all show up as `None` for `natom = 1` molecules.summary documents. Think we did this because `float("inf")` isn't directly JSON serializable but has custom BSON types

Added a bit of validation logic to remove these values consistent with what we have in the database

PR #1435, opened 2026-04-20
emmet 0.87.0 tweaks

esoteric-ephemera:
Changes required for full compatibility with data + model changes:
- `oxi_states` --> `oxidation_states`collection name
- XAS spectrum IDs now use single enum rather than `list[Enum]`. Will require data rewrite
- Aggregation pipeline for `SuffixedID` fields (thermo ID, battery ID, and XAS spectrum ID)


esoteric-ephemera:
@tsmathis with the [repr revert](https://github.com/materialsproject/emmet/pull/1435/changes/571f8ba9dd9a7258b94e492963f2e96bb03a528f): Didn't we want the repr to be MPID for legacy IDs and AlphaID otherwise?

tsmathis:
Yes (-ish, I think I was off base with my intentions around that though - introducing too much complexity)

Reverted though due to this test breaking here: [test_alpha_Id.py:99-103](https://github.com/materialsproject/emmet/blob/554cd3fca350d6ff02383f45f8d173cd2a0d9468/emmet-core/tests/test_mpid.py#L99-L103)

checking out [10b55cb](https://github.com/materialsproject/emmet/pull/1435/commits/10b55cbddfdb97050827102a7ea7434f5d35837d) you'll get this:

```
>>> ref_idx = AlphaID(137, prefix="cats", separator=":")
>>> ref_idx
MPID(cats-137)
```
so 1) something was a little bit off there, which lead me to 2) concluding that it's probably not worth it due to the added complexity. 

Will also invoke the phrase "principle of maximum surprise" again for instantiating a class and getting a different class in the repr. So my bad on that one, should have thought that through better.

esoteric-ephemera:
OK sure I see what you mean - I think we'll have a sense of if this needs to change again once we have a test deployment ready

PR #1439, opened 2026-04-23
NEP 29 support - preparing for py3.12+

esoteric-ephemera:
Numpy has now dropped support for py3.11 as of 24 April, 2026. Keeps support for py3.11 (not in the spirit of [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)), but adds requirements + testing for py3.12 - 3.14


tsmathis (approved):
LGTM

PR #1445, opened 2026-05-04
misc updates ahead of 0.87.0 rc

tsmathis:


emmet-api/emmet/api/routes/materials/xas/query_operators.py
```diff
@@ -34,10 +32,35 @@ def ensure_indexes(self):  # pragma: no cover
...
+            ]
+
+            fields = {
+                "material_id": processed_identifiers,
+                "spectrum_type": list(set(spectrum_types)),
+                "absorbing_element": list(set(absorbing_elements)),
+                "edge": list(set(edges)),
+            }
+
+            crit = {k: {"$in": v} for k, v in fields.items()}
```
esoteric-ephemera: Isn't this missing a few steps in aggregation? This will likely give more results than it should, since it needs to narrow on which IDs were initially queried. Like if the user requests 
`spectrum_ids = ['mp-149-XANES-Si-K','mp-13-XANES-Fe-L']` , 
then all of the `L` edge results would also get returned for `mp-149` and all the `K` results for `mp-13`

tsmathis: Yep, it will. But I don't really see another way to utilize `material_id`, `spectrum_type`, `absorbing_element`, and `edge` over `spectrum_id`.

Should likely just go back to `SuffixedIDQuery` then? Remind me, was the compound `XasLabel` enum already working with that query operator? Or was it needing to be modified?

tsmathis: The other option is of course to just compute the spectrum_id and add it back as a concatenated string in the db?

esoteric-ephemera:
Notes for us later:
- Conversion electrodes will need a legacy str match search on their `battery_id` field. There's no `material_ids` field to use in searching right now, unlike for insertion electrodes. Will wait on a solution for the `task_id`-like query to proceed on this

tsmathis:
> Conversion electrodes will need a legacy str match search on their battery_id field. There's no material_ids ...

Okay, will definitely need to update the data for this one then

esoteric-ephemera:
@tsmathis refactored the compound ID (formerly `SuffixedID`) stuff heavily per our chats:
- `emmet.core.types.typing.validate_compound_identifier` takes over the role of validating compound IDs
  - In `emmet.core.thermo`, `electrodes`, and `xas`, there are specific wrapped versions of this function with the suffixes pre-defined. These take over the role of subclasses of `SuffixedID`
- `emmet-api` reuses those specific validators to parse a list of compound IDs. The API only does an `$in` on each set of unique ID components, so users will get more results than anticipated
  - No post-processing on these

Format changes, think I got these right but lmk and I'll modify accordingly:
- `thermo_id`: "mp-abcdefgh-ThermoType"
- `battery_id`: "mp-abcdefgh_Element"
- `spectrum_id`: "abcdefgh-XasType-Element-XasEdge" (since these are task IDs)

esoteric-ephemera (approved):

PR #1454, opened 2026-05-12
pymatgen I/O layer

esoteric-ephemera:
- Define an I/O layer between emmet-core and pymatgen. Inspired by [this pymatgen pr](https://github.com/materialsproject/pymatgen/pull/4633) post reorg
  - Ensures we only have one point of breakage if pymatgen/-core changes import statements
  - Based around lazy loading of pymatgen modules
- Clean up tests to only construct test data once


esoteric-ephemera:
@tsmathis broke the dependencies in the I/O layer into separate `pymatgen-core`, `pymatgen`, and `pymatgen`-add-on (like `pymatgen-analysis-alloys`) sets [here](https://github.com/materialsproject/emmet/blob/6021e54a9b2d3f0f4fa213a544ee84a01a936c75/emmet-core/emmet/core/io/pymatgen.py#L12)

To me, it looks infeasible at this time to separate out non-core deps from emmet-core

The I/O layer itself is still gtg

esoteric-ephemera:
@tsmathis to the nightly testing question: we can probably do something minimal with
`pytest emmet-core/tests/io/test_pymatgen.py::test_imports`
which will test all the import

PR #1490, opened 2026-07-15
CLI: invalidate cached validation when any file changes

### LINKED ISSUE #1485: [Bug]: emmet-cli can retain stale validation after a file changes
### Code snippet

```python
calculation.calc_valid = True

# Change a file that is not the final item in calculation.files.
calculation.files[0].path.write_text("changed")
calculation.refresh()

assert calculation.calc_valid is None
```

### What happened?

`CalculationMetadata.refresh()` overwrites `changed_files` during every loop iteration:

```python
changed_files = cached_hash != f.hash
```

As a result, only the final file determines whether cached validation is cleared. If an earlier file changes and the last file does not, `calc_valid` and `calc_validation_errors` can remain stale.

Expected behavior: changing any calculation file invalidates cached validation.

Proposed solution: accumulate the result across all files, for example with:

```python
changed_files |= cached_hash != f.hash
```

Acceptance criteria:

- Any changed file clears cached validation.
- An unchanged calculation retains its cached result.
- Tests cover first, middle, and final file changes.

kbuma:
## Summary

- accumulate file-change detection across every file in `CalculationMetadata.refresh()`
- invalidate cached validation when the first, middle, final, or multiple files change
- preserve cached validation and errors when every file is unchanged

Closes #1485

## Contributor Checklist

- [x] I have broken down my PR scope into the following TODO tasks
  - [x] fix accumulated file-change detection
  - [x] cover changes at every file position
  - [x] cover simultaneous changes and the unchanged case
- [x] I have run the tests locally and they passed.
- [x] I have added tests, or extended existing tests, to cover the bug fixed in this PR.

## Testing

- `pytest -q emmet-cli/tests` — 71 passed, 2 skipped
- `mypy --config-file .mypy.ini emmet-cli/emmet` — passed
- `pre-commit run --all-files --show-diff-on-failure` — passed

### DIFF
--- a/emmet-cli/tests/test_submission.py
+++ b/emmet-cli/tests/test_submission.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from emmet.cli.submission import CalculationMetadata, CalculationLocator, Submission
 from emmet.cli.utils import EmmetCliError
+from emmet.core.vasp.utils import FileMetadata
 import pytest
 
 
@@ -35,6 +36,23 @@ def tmp_structure(tmp_path_factory):
         else:
             tmp_structure[calc_dir] = a_file
     return tmp_structure
+
+
+@pytest.fixture()
+def calculation_metadata(tmp_path):
+    files = []
+    for index in range(3):
+        file_path = tmp_path / f"file-{index}"
+        file_path.write_text(f"initial content {index}")
+        file_metadata = FileMetadata(name=file_path.name, path=file_path)
+        file_metadata.compute_hash()
+        files.append(file_metadata)
+
+    return CalculationMetadata(
+        files=files,
+        calc_valid=True,
+        calc_validation_errors=["cached validation error"],
+    )
 
 
 def verify_submission_calculations_against_tmp_dir_data(calculations):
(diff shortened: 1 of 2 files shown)


emmet-cli/tests/test_submission.py
```diff
@@ -37,6 +38,23 @@ def tmp_structure(tmp_path_factory):
...
+        file_path = tmp_path / f"file-{index}"
+        file_path.write_text(f"initial content {index}")
+        file_metadata = FileMetadata(name=file_path.name, path=file_path)
+        file_metadata.compute_hash()
+        files.append(file_metadata)
+
+    return CalculationMetadata(
+        files=files,
+        calc_valid=True,
+        calc_validation_errors=["cached validation error"],
```
tsmathis: Forgive me if I'm just missing something obvious, but this looks like the opposite of what it should be? If `calc_valid=True`, shouldn't `calc_validation_errors=[]`?

similar questions below

emmet-cli/tests/test_submission.py
```diff
@@ -149,6 +167,48 @@ def test_changed_files(sub_file):
...
+def test_refresh_invalidates_cached_validation(calculation_metadata, changed_index):
+    original_hashes = [file.hash for file in calculation_metadata.files]
+    changed_file = calculation_metadata.files[changed_index]
+    changed_file.path.write_text("changed content")
+
+    calculation_metadata.refresh()
+
+    assert changed_file.hash != original_hashes[changed_index]
+    assert calculation_metadata.calc_valid is None
+    assert calculation_metadata.calc_validation_errors == []
```
tsmathis, line 180: related to first comment, based on the test name seeming falsey, I would expect  `calculation_metadata.calc_validation_errors` to not be an empty list? Same applies to next test for multiple changed files

emmet-cli/tests/test_submission.py
```diff
@@ -149,6 +167,48 @@ def test_changed_files(sub_file):
...
+
+
+def test_refresh_preserves_cached_validation_for_unchanged_files(calculation_metadata):
+    original_hashes = [file.hash for file in calculation_metadata.files]
+
+    calculation_metadata.refresh()
+
+    assert [file.hash for file in calculation_metadata.files] == original_hashes
+    assert calculation_metadata.calc_valid is True
+    assert calculation_metadata.calc_validation_errors == ["cached validation error"]
```
tsmathis: related to first comment, based on the test name I would expect there to no be no `calc_validation_errors`

tsmathis (commented):
all the comments are related, so I might just be missing something.
is setting the fixture to be: 
```python
        ...
        calc_valid=True,
        calc_validation_errors=["cached validation error"],
        ...
```
just to carry those values through to check if they changed?
Rather than to be representative of the state that an actual valid `CalculationMetadata` would be? 
(would expect this for example in a valid calc meta)
```python
CalculationMetadata(
        files=files,
        calc_valid=True,
        calc_validation_errors=[],  # no errors
)
```

tsmathis:
if the tests are just checking for state to be preserved (or flipped) then the actual values don't matter and all my comments above can be resolved

emmet-cli/tests/test_submission.py
kbuma: I missed this addition by Codex. it should be that - fixed

emmet-cli/tests/test_submission.py
kbuma: this is actually correct; when a refresh happens and the cache is no longer valid everything gets cleared out until validation is run again

emmet-cli/tests/test_submission.py
kbuma: yup - fixed

kbuma:
@tsmathis thanks for the catch; fixed now

tsmathis (approved):

PR #1503, opened 2026-08-12
add _get_number_to_space_group_symbol_mapping

minhsueh:
Problem:
The mapping from space group international number to symbol is one-to-many.
This function picks the latest entry in `SYMM_DATA["abbreviated_spacegroup_symbols"]`
to produce a symbol consistent with `symmetry.symbol` used in MP, rather than the default
pymatgen symbol (`from pymatgen.symmetry.groups import sg_symbol_from_int_number`). The affected groups (monoclinic, unique axis b vs. full
symbol) are:

international number | symbol in MP | symbol in PMG
--------------------- | ------------- | --------------
15                     | C2/c          | C12/c1
12                     | C2/m          | C12/m1
5                      | C2            | C121
9                      | Cc            | C1c1
8                      | Cm            | C1m1
13                     | P2/c          | P12/c1
10                     | P2/m          | P12/m1
3                      | P2            | P121
14                     | P2_1/c        | P12_1/c1
11                     | P2_1/m        | P12_1/m1
4                      | P2_1          | P12_11
7                      | Pc            | P1c1
6                      | Pm            | P1m1

Also, 4 orthorhombic groups with legacy vs. current e-glide symbol:
international number | symbol
--------------------- | ------------- 
39                     | Abm2/Aem2
41                     | Aba2/Aea2
67                     | Cmma/Cmme
68                     | Ccca/Ccce

Note: relying on "the latest entry" depends on Python dict insertion/iteration
order, which is not ideal practice, but it works for now.


tsmathis (commented):
> Note: relying on "the latest entry" depends on Python dict insertion/iteration
order,

Can this drift over time then? 
Are there any sane tests that can be added to prove this produces the symbols we want? (and therefore detect drift)

tsmathis:
and this is or is not related to #1501 ?

minhsueh:
@tsmathis 
- Tests added
- Add reason to `xfail` from [this PR](https://github.com/materialsproject/emmet/pull/1501)

About this (although we've discussed this earlier face-to-face, just for future reference):
> and this is or is not related to https://github.com/materialsproject/emmet/pull/1501 ?

No — PR 1501 is about parsing a string into the correct form (subscripts and bars). 
This one is about mapping space group numbers to symbols, which is one-to-many (due to different standards), and this function is specifically about returning the symbols that MP uses.

tsmathis (approved):
tests should pass after running deps upgrade

PR #1513, opened 2026-08-31
cli: harden task state handling

kbuma:
This PR fixes flaky CLI background-task tests and hardens task state updates against parent/child races. It supersedes closed PR #1512 and is rebased onto the current main branch.

- Atomically merge task-state updates under the state-file lock.
- Preserve the first terminal task status while still accepting late PID metadata.
- Replace timing sleeps with bounded state polling.
- Use synthetic task state for mocked process scenarios and clean up live tasks during teardown.
- Preserve main's explicit fork multiprocessing context for Python 3.14.

## Contributor Checklist

- [x] I have broken down my PR scope into the following TODO tasks
   - [x] Harden concurrent task-state updates
   - [x] Remove race-prone process test setup
   - [x] Rebase onto current main and resolve conflicts
- [x] I have run the tests locally and they passed.
- [x] I have added tests, or extended existing tests, to cover the bugs fixed in this PR.

## Testing

- pre-commit run --all-files
- Mypy: passed for emmet-cli/emmet
- CLI tests with coverage: 78 passed, 1 skipped

The local pytest environment did not include pytest-xdist, so the suite ran serially rather than with -n auto.

### DIFF
--- a/emmet-cli/emmet/cli/state_manager.py
+++ b/emmet-cli/emmet/cli/state_manager.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import logging
 from pathlib import Path
-from typing import Any, Self, TextIO
+from typing import Any, Callable, Self, TextIO
 import fcntl
 
 logger = logging.getLogger("emmet")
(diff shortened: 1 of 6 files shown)


tsmathis (approved):
testing is only failing due to some upstream dependency conflicts messing with tests.
will fix elsewhere

